PR #32, opened 2018-12-10
update to OpenAPI 3.0

patrickkwang:
In addition to the version update, there is just one _substantive_ change to the specification, filling out response types for:
* /feedback/ratings
* /feedback/expertise_levels
* /result/{result_id}/feedback

A few examples have also been fixed to actually fit the spec.

### DIFF
--- a/API/0.9.0/TranslatorReasonersAPI_0.9.0.yaml
+++ b/API/0.9.0/TranslatorReasonersAPI_0.9.0.yaml
@@ -70,6 +70,8 @@ paths:
       responses:
         200:
           description: "successful operation"
+          schema:
+            $ref: "#/definitions/Ratings"
       x-swagger-router-controller: "swagger_server.controllers.feedback_controller"
   /feedback/expertise_levels:
     get:
@@ -83,6 +85,8 @@ paths:
       responses:
         200:
           description: "successful operation"
+          schema:
+            $ref: "#/definitions/ExpertiseLevels"
       x-swagger-router-controller: "swagger_server.controllers.feedback_controller"
   /feedback/all:
     get:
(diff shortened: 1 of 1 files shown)


edeutsch:
Apologies for the delay in responding to this. This mostly looks fine to me. My OpenAPI parser seems to be unhappy about having a simple type under a schema section. So the constructions like this:
  name: result_id
  description: Integer identifier of the result to return
  required: true
  schema:
    type: integer
cause a parsing error:
main] WARN io.swagger.codegen.DefaultCodegen - Type is NULL for Serializable Parameter: result_id
[main] ERROR io.swagger.codegen.DefaultCodegen - String to be sanitized is null. Default to ERROR_UNKNOWN

But if I change it to:
  name: result_id
  description: Integer identifier of the result to return
  required: true
  type: integer
then it is happy.

I would be fine to merge the pull request and then consider making this change if it is not detrimental to other uses of the file.  Or we can leave it as is, and I can just make local deltas for our use. It is semantically the same I think.

What do you think?

patrickkwang:
My reading of the OpenAPI spec is that `type` is not a standard field for the parameter object, and should be encased in a `schema`.
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#parameter-object

Is it possible that your Codegen is out of date?
https://github.com/swagger-api/swagger-codegen#compatibility

edeutsch:
ah, now I've figured it out. I was using the latest. But the latest from the 2.x branch. I've now figured out how to pull the latest 3.x branch, which is great. But now that I've got it running, I don't see any way of generating Python code. only support languages are:
aspnetcore
csharp
csharp-dotnet2
dynamic-html
html
html2
java
jaxrs-cxf-client
jaxrs-cxf
inflector
jaxrs-cxf-cdi
jaxrs-spec
jaxrs-jersey
jaxrs-di
jaxrs-resteasy-eap
jaxrs-resteasy
spring
nodejs-server
openapi
openapi-yaml
kotlin-client
kotlin-server
php
scala
scala-akka-http-server
swift3
swift4
typescript-angular

bummer.

PR #33, opened 2019-03-04
Alternative: embedded vs. remote knowledge graph

patrickkwang:
This was implicit before, but can be made explicit now using `oneOf` with OpenAPI 3.0.


edeutsch:
These changes obsoleted due to repo reorg.

PR #36, opened 2019-03-05
Requirements

patrickkwang:
A number of fields are effectively required; reasoning cannot proceed without them. Let's make these explicit.


patrickkwang:
Yes, robokop does the same, but expects an empty list to be provided for `edges` in those cases. Formally, a graph consists of a set of vertices and a set of edges, even if they are empty sets. Practically, it will be more concise to just iterate over an empty list than to check whether the `edges` field exists. So my inclination is to require that `edges` exists, but put no requirements on its length.

PR #46, opened 2019-03-26
Fix for OpenApi code generation

lhannest:
These changes will allow [Connexion](https://github.com/zalando/connexion/issues) to pass the request body into the [generated](https://github.com/OpenAPITools/openapi-generator) controller methods for the two post methods.

Unfortunately OpenAPI Generator is a bit inconsistent in how it chooses parameter names. Having `additionalProperties: true` in `Query` causes it to not use the lowercase name of the object (in this case "query") as the parameter name and instead default to "request_body". Otherwise, when this property isn't set, it seems to choose the lowercase name of the object (as can be seen in the case of `Feedback`, where "feedback" is chosen as the parameter name).

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -283,6 +283,7 @@ components:
         tag:
           type: string
     Query:
+      x-body-name: request_body
       type: object
       properties:
         bypass_cache:
@@ -910,6 +911,7 @@ components:
           description: A URL corresponding to this attribute
       additionalProperties: true
     Feedback:
+      x-body-name: feedback
       type: object
       description: A single unit of Feedback
       properties:


lhannest:
To try using the code generator:
```
wget http://central.maven.org/maven2/org/openapitools/openapi-generator-cli/3.2.3/openapi-generator-cli-3.2.3.jar -O openapi-generator-cli.jar
java -jar openapi-generator-cli.jar generate -i TranslatorReasonersAPI.yaml -g python-flask -o server
```
The generated README.md explains how to run it.

edeutsch:
Yes, this is what I tried (although via the docker version of the same thing). It generates the code I need, but when I launch it, it poops out with a puzzling error message that I gave up trying to fix. Seems like it almost there, but not quite.

edeutsch:
Ah, but I didn't try these new lines in the pull request. Will see if that helps.

edeutsch:
Just to confirm, the addition of:
     x-body-name: request_body
was indeed the trick I needed to get rid of the mysterious problem that I could not solve. Thank you. The puzzling error message was:
TypeError: query() missing 1 required positional argument: 'request_body'

PR #47, opened 2019-03-29
make query_graph objects more visible

RichardBruskiewich:
... in the README, in the context of their use as a possible input parameter.


API/README.md
```diff
@@ -31,9 +31,8 @@ is provided and decide if it can proceed with the input given.
...
-
+- **query_message** - object - Message object that represents the query to be answered (e.g.: ) which contains a [query graph representing the structure of a query](https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI/blob/master/API/TranslatorReasonersAPI.yaml#L664).                                                                                           
+- **previous_message_processing_plan** - object - Container for one or more Message objects or identifiers for one or more Messages along with a processing plan for how those messages should be processed and returned (e.g.: ). Contains embedded **Message** objects similar to the ones referenced by **query_message** above), which also contain a [query graph object representing the structure of a query](https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI/blob/master/API/TranslatorReasonersAPI.yaml#L664)
```
patrickkwang: There may be a missing parenthesis here: `(similar...`

RichardBruskiewich: Fixed. Took the opportunity to move things around a bit.

edeutsch:
From the RTX perspective a query_message can contain a Message with *either* a query_graph or a query_type_id & terms (i.e. a "canned query"). Maybe worth adding that if we want to be more precise.

And for the previous_message_processing_plan, the key difference is that these are Message object(s) that already contain *answers* that should be operated on. So, while these Message objects may well contain a query_graph, they should also contain a knowledge_graph and results.

This is perhaps a little silly to separate them out this way. Maybe a better design would have been one way to send messages in, irrespective of whether they are answered or not, and then the reasoner engine would have to examine each one and decide what to do. If the Message already includes an answer, then that answer can be accepted for further processing; if no answer, then it is a question to be asked. Something to consider for 1.0

PR #50, opened 2019-04-01
clean up storage endpoints

patrickkwang:
These message/result/feedback storage mechanisms are independent of the main reasoner service provided by `/query` and are not required to pipeline modules.


edeutsch:
it is okay with me to remove the feedback mechanisms (although if anyone else is thinking of implementing it, why not leave it?)

BUT, I think it is fairly important to leave at least the /message/{id} mechanism
I think also the /result/{id} mechanism would be useful although one might argue that with the new scheme, the results are extremely dependent on their parent message that maybe having identifiers is less important? I'm still hesitant to remove this unless I'm the last voice in the forest calling for this.

patrickkwang:
My feeling is that this standard should be minimal in the sense that it only specifies components and endpoints that are required to make reasoner modules interoperable. Some kind of storage functionality is useful for web GUIs (ROBOKOP does it too, differently) and can be implemented independently from the standardized `/query` endpoint. But unless we're proposing to attempt asynchronous REST requests with polling, I don't think it is useful to a reasoner pipelining tool.

patrickkwang:
Per today's Webex discussion, the feedback/storage endpoints will be moved into a separate ("maximal") file as recommendations for those who wish to implement them, while the `/query` endpoint will remain in the "required" specification. A pull request to this effect is forthcoming.

PR #60, opened 2019-05-16
minimal requirements for reasoner support

patrickkwang:
See https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI/issues/29#issuecomment-492408988.
Here is a suggestion for the components/fields that _all_ reasoners should support.

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -49,195 +49,15 @@ components:
       x-body-name: request_body
       type: object
       properties:
-        bypass_cache:
-          type: string
-          example: 'true'
-          description: >-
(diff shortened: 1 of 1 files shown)


patrickkwang:
The issue, as I tried to explain in https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI/issues/29#issuecomment-492408988, is that we can't distinguish between the elements that the user is required to provide, and those that the server is required to support. We could, for instance, require reasoners to support `max_results`, without requiring users to provide it. An individual OpenAPI document must address the requirements on the user, while to indicate that something is optional for any server implementation, we're using this awkward two-file system.

Issue #29, opened 2018-11-19
Required fields


patrickkwang:
Version 0.9.0 includes multiple ways to transmit information:
* in `Message`, `query_type_id` and `terms` vs. `query_graph`
* in `Result`, `result_graph` vs. `knowledge_map`

There are also a lot of other fields that may or may not be desired or provided by different reasoners.

In order to build a reasoner module that implements this format, or an algorithm to call such modules, we need to know which elements are required. Reasoner modules must implement at least the required elements, and the meta-algorithm should be okay in cases where anything not-required is missing.

In cases where we'd like to require `A` OR `B`, OpenAPI 3.0 provides the `oneOf` keyword (I'm not sure if this is available in 2.0). Using this structure, `Message` could, for example, require `query_type_id` AND `terms` OR `query_graph`. This may not be ideal for interoperability purposes, but it would give us a place to start.

edeutsch:
I was thinking that this might be best left to some additional documentation and policies and letting the code make decisions based on what is available, rather than prescribing requirements in the YAML. One could even see some benefits to letting a reasoner output both styles as we learn how to support the new format. But I don't feel strongly. If you'd like to propose which fields you think should be required, we could have those discussions.

patrickkwang:
We've now split the specification into two versions: a 'minimal' version that every reasoner ought to support, and an 'optional' version providing standards for additional properties/functionality (e.g. message/feedback storage mechanisms).
This is getting at a problem that we've always had and never really addressed. Namely, an OpenAPI document describes a specific API rather than a class of APIs. When the OpenAPI documentation indicates that an input parameter is optional, that means it's optional for the client of that API, but the server will always handle the parameter if it is provided. To specify that different servers may support different sets of optional parameters, we need these two versions of the OpenAPI document.
All of this is to illustrate that indicating that certain fields in e.g. [Query](https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI/blob/master/API/TranslatorReasonersAPI.yaml#L48) are required is not sufficient, because even 'optional' properties like `asynchronous` must be supported by the server. The solution is to remove these fields entirely from the minimal OpenAPI document, indicating that not all reasoners must support them. They will remain in the optional OpenAPI document to provide standardization between reasoners that do intend to support them.
Each reasoner's individual OpenAPI document will indicate which parameters and fields it supports.

PR #61, opened 2019-05-18
misc requirements

patrickkwang:
Query: query_message
Result: node_bindings and edge_bindings
QNode: node_id
QEdge: edge_id

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -55,6 +55,8 @@ components:
           items:
             $ref: '#/components/schemas/Message'
       additionalProperties: true
+      required:
+        - query_message
     Message:
       type: object
       properties:
@@ -105,6 +107,9 @@ components:
               - type: array
                 items:
                   type: string
+      required:
+        - node_bindings
+        - edge_bindings
     KnowledgeGraph:
       type: object
       description: >-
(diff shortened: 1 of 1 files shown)


patrickkwang:
#29

edeutsch:
For us, a query_message{} is not required if a previous_message_processing_plan{} is provided.

edeutsch:
also, for us, not 100% of Results will have node_bindings and edge_bindings.

I do not think that enforcing this minimal set of requirements will benefit us. Most of our query results will conform to this, but I know that we do currently and will have some results that cannot conform to these requirements.

Is it better that we set very restrictive requirements knowing that some reasoners won't always conform to these requirements?

Or is it better to leave these things somewhat unconstrained?

I can go ahead and approve this if you like and we can merge it, but RTX will sometimes violate it.

patrickkwang:
The purpose of this standard is to promote/ensure interoperability. The bindings format is critical for allowing us to evaluate, compare, and build upon reasoner results. There will of course be services that cannot conform to the standard and therefore won't be used by whatever uber-reasoner aggregates or pipelines these services. But it sounds like you have a single endpoint that only sometimes conforms to the this format? Is there a chance this can be split into multiple endpoints, one of which is fully compliant with the proposed spec?

vdancik:
For us `qg_id` is always identical to `kg_id`. I think node_binding and edge_binding should be optional for cases when `qg_id != kg_id`. The remaining three requirements (query_message, node_id, edge_id) are not problematic for us.

edeutsch:
I think this PR is obsolete. At minimum, the attribute names have changed by now. Can we expunge this PR?

patrickkwang:
This branch is once again fully ahead of master.

PR #62, opened 2019-05-20
Add recommendation to use biolink-model for entity type

cmungall:


patrickkwang:
Good idea. Should we add an `externalDocs` for these pointing to some official resource for the biolink model?
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields-20

patrickkwang:
If we make `BiolinkNode` and `BiolinkEdge` components, these could contain the external documentation links and reduce some redundancy.

patrickkwang:
The way I read this, it seems like we want a [related_to](https://biolink.github.io/biolink-model/docs/related_to.html) rather than an [edge_label](https://biolink.github.io/biolink-model/docs/edge_label.html).

edeutsch:
I confess that I do not understand these last two comments.

patrickkwang:
@cmungall: Do you follow? Do you agree? Would you like to make the edits?

patrickkwang:
@cmungall I can't seem to add you as a reviewer, but please let us know if this looks reasonable.

Issue #63, opened 2019-05-20
Implement Entity Relations Schema


MarkDWilliams:
Implementing a schema for each reasoner API which describes the entity types present, which entities they can be linked to, and by which predicates.  Also adding an API endpoint which returns this information.  The format is similar to the one currently implemented by ROBOKOP which is described below.
` 
schema:

    type: object

    additionalProperties:

        type: object

        additionalProperties:

            type: array

            items:

                type: string`

PR #64, opened 2019-05-20
/predicates endpoint

patrickkwang:
For comparison, here is the alternative proposal:
https://biomedical-translator.slack.com/archives/C92KXC9DL/p1557167017024100

The "array of triples" would be much more flexible, allowing addition of metadata (e.g. number of such edges, other source/target attributes?) without structural changes, and potentially avoiding future issues of the type we've had in other places (https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI/issues/56#issuecomment-489767605).

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -28,10 +28,10 @@ paths:
           content:
             application/json:
               schema:
-                description: Target map
+                description: Source map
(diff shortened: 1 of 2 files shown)


patrickkwang:
#63

edeutsch:
Also need to add this new section to the _optional.yaml too?

PR #65, opened 2019-05-20
more flexible bindings format

patrickkwang:


edeutsch:
The structure seems fine to me. However, everywhere else in the format we've opted for longer and descriptive names, so I would discourage using cryptic short names here (qid, kid). How about we use longer and descriptive names here, too? like query_graph_node_id and knowledge_graph_node_ids (presumably plural because it can be plural? Or if greater brevity seems important, at least qg_node_id and kg_node_ids?

patrickkwang:
These little bindings objects can occur tens of thousands of times in a single message, so I'm very much in favor of brevity for the sake of limiting the use of network resources. These will always occur under the heading of `node_bindings` or `edge_bindings`, so I don't feel the need to state again that these are node/edge ids.

How about `qg_id` and `kg_id`?

patrickkwang:
Re: multiple KG ids, we could with this structure insist that they all be independent binding objects. This will be required anyway in cases where different metadata is attached to different bindings to the same QG node/edge.

For example:
```
[
  {
    'qg_id': 'n00',
    'kg_id': 'MONDO:0005737',
    'score': 0.9
  },
  {
    'qg_id': 'n00',
    'kg_id': 'MONDO:0004979',
    'score': 0.72
  }
]
```

edeutsch:
okay, I am fine with qg_id and kg_id.

In your second comment, are you suggesting to change the schema so that kg_id is always scalar and not array?

patrickkwang:
Yes. It's not necessary, but it's now possible, and any simplification like that should make life easier for folks implementing the spec. Regardless, we can put off that question for another day, another pull request.

PR #67, opened 2019-07-03
remove redundant prefixes

patrickkwang:


patrickkwang:
There are two main reasons to do this:
1. All other things being equal, a more concise message is better.
2. Stylistically, we'd like to avoid "stuttering": `query.query_message`, `qnode.node_id`, `qedge.edge_id`

The `description`s of these properties exist to clarify how they are to be used.

PR #83, opened 2019-08-14
Update TranslatorReasonersAPI.yaml

edeutsch:


API/TranslatorReasonersAPI.yaml
```diff
@@ -13,6 +13,11 @@ externalDocs:
   description: Documentation for the NCATS Biomedical Translator Reasoners web services
   url: 'https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI'
 tags:
+  - name: predicates
```
patrickkwang: I like this. We should attach this tag to the /predicates endpoint.

edeutsch: ah, thanks, forgot about that. Now added.

PR #87, opened 2019-09-25
allow str or list(str)

### LINKED ISSUE #86: str vs. list(str)
There are a number of places where this should be addressed:

- query_graph.nodes[].curie
- query_graph.nodes[].type
- query_graph.edges[].type
- results[].node_bindings[].kg_id
- knowledge_graph.nodes[].type

Probably more. My first instinct is to allow any of these to be either a string or list of strings, with some shared understanding of what it means to have a list for e.g. a query-graph node type.

patrickkwang:
in:
* query_graph.nodes[].curie
* query_graph.nodes[].type
* query_graph.edges[].type
* results[].node_bindings[].kg_id
* knowledge_graph.nodes[].type

Resolves #86

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -247,11 +247,19 @@ components:
             QueryGraph internal identifier for this QNode. Recommended form: n00,
             n01, n02, etc.
         curie:
-          type: string
+          oneOf:
+            - type: string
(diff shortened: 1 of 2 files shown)


patrickkwang:
Do you mean the issue #86? These are exactly the modifications I suggested there. Maybe I should have gone straight for the pull request.

PR #89, opened 2020-01-14
Update TranslatorReasonersAPI.yaml

RichardBruskiewich:
Clarify  description for Edge schema:   local id pertains to an edge, not a node?

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -321,7 +321,7 @@ components:
           type: string
           example: '553903'
           description: >-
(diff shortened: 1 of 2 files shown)


RichardBruskiewich:
ok.  might be nice to have one point of changes for the spec (somehow) with exceptions  generated by  some automatic process.

patrickkwang:
I agree, this is very awkward. This is the system we arrived at for denoting which components of the API are required as well as imposing constraints on some of the optional components. Perhaps a more elegant (and enforceable) way to describe this standard would be a JSON Schema to augment the OpenAPI Specification.

PR #100, opened 2020-04-08
remove RemoteKnowledgeGraph

patrickkwang:


patrickkwang:
There is one ROBOKOP [microservice](http://robokop.renci.org:4868/ui/#/reasoner/answer) that uses this functionality. To the best of my knowledge, it is only used internally by ROBOKOP. ROBOKOP's main 0.9.2 [endpoint](https://robokop.renci.org/ranker/apidocs/#/answer/post_ranker_api_query_) never returns a RemoteKnowledgeGraph. I'm happy to do any necessary internal refactoring.

But I think the more important question is, is RemoteKnowledgeGraph a generally useful feature? I tried to figure out where it first originated, and all I found was [this](https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI/commit/26e17afcfdae46afdb838f6ab8677d368dccea42#diff-f220f224d5663a927a737c9b146c30efR384). I think the main idea was to save bandwidth by allowing result bindings to reference a remote database, rather than package all node/edge data into the Message. This would allow clients to then go and extract only the data that they need. As an aside, it occurs to me that [GraphQL](https://neo4j.com/developer/graphql/) could be a great way to serve this data.

The RemoteKnowledgeGraph functionality has been around for a while, and the lack of adoption indicates to me that a) it's not a necessary or desirable feature or b) our documentation/marketing are insufficient. Probably both.

An alternative that could serve the same purpose is to allow a null knowledge_graph, but require that all node (and edge?) ids be CURIEs that resolve to a publicly accessible URLs (e.g. [MONDO:0100096](http://purl.obolibrary.org/obo/MONDO_0100096)) with machine-readable data, somehow.

Issue #94, opened 2020-02-10
Use objects instead of lists for *graph nodes and edges


patrickkwang:
Anecdotally, when parsing a query_graph or knowledge_graph in Python, I almost always want a map from node/edge ids to the corresponding details, and I have to construct this whenever I receive a new *graph. If the native format were that map instead of the current list, it would be handy.

This change would also make enforceable the currently unstated requirement that node/edge ids must be unique.

edeutsch:
yeah, I agree, this was my fault from long ago. I think there are a couple other places where we should reconsider lists like node_attributes[] and edge_attributes[]. Anyway, I assume there'll be no action on this before the March demos, but I would highly encourage that we reconsider this in April. This will be a huge breaking change everywhere if we do change it. But it probably makes sense to do.

patrickkwang:
#101

PR #101, opened 2020-04-09
Convert *graph nodes/edges and *bindings to maps

patrickkwang:
This is a draft proposal to address #94. It has three parts:
1. Convert `query_graph` `nodes` and `edges` fields from lists to maps. The `id` field of each becomes the associated key.
2. Convert `knowledge_graph` `nodes` and `edges` fields from lists to maps. The `id` field of each becomes the associated key.
3. Convert `node_` and `edge_bindings` from lists to maps. The `qg_id` field of each becomes the associated key. The value is a list of all bindings using that `qg_id` because, unlike with 1 and 2, the `qg_id`s are not expected to be unique (i.e. there may be multiple entities bound to each query element).

These changes come with some benefits:
* This restructuring is often performed by servers/clients anyway, so we can ease the burden on them.
* We can enforce uniqueness of `id`s in both `query_` and `knowledge_graph`s.

I also have some concerns:
* This will require a significant refactoring effort for current implementers.
* [re: 2] In the `knowledge_graph`, `id`s are typically of the form 'XXX:YYY', and are not valid attribute names in Python or JavaScript. This will not prevent the use of these structures, but may limit some options.

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -114,14 +114,14 @@ components:
       description: One of potentially several results or answers for a query
       properties:
         node_bindings:
-          type: array
+          type: object
           description: List of QNode-KNode bindings.
-          items:
+          additionalProperties:
             $ref: '#/components/schemas/NodeBinding'
         edge_bindings:
-          type: array
+          type: object
           description: List of QEdge-KEdge bindings.
-          items:
+          additionalProperties:
             $ref: '#/components/schemas/EdgeBinding'
       required:
         - node_bindings
@@ -129,9 +129,6 @@ components:
     NodeBinding:
       type: object
       properties:
-        qg_id:
-          type: string
-          description: Query-graph node id, i.e. the `node_id` of a QNode
         kg_id:
           oneOf:
             - type: string
@@ -140,14 +137,10 @@ components:
                 type: string
           description: One or more knowledge-graph node ids, i.e. the `id` of a KNode
       required:
-        - qg_id
         - kg_id
     EdgeBinding:
       type: object
       properties:
-        qg_id:
-          type: string
-          description: Query-graph edge id, i.e. the `edge_id` of a QEdge
         kg_id:
           oneOf:
             - type: string
@@ -156,7 +149,6 @@ components:
                 type: string
           description: One or more knowledge-graph edge ids, i.e. the `id` of a KEdge
       required:
-        - qg_id
         - kg_id
     KnowledgeGraph:
       type: object
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -117,12 +117,16 @@ components:
           type: object
           description: List of QNode-KNode bindings.
           additionalProperties:
-            $ref: '#/components/schemas/NodeBinding'
+            type: array
+            items:
+              $ref: '#/components/schemas/NodeBinding'
         edge_bindings:
           type: object
           description: List of QEdge-KEdge bindings.
           additionalProperties:
-            $ref: '#/components/schemas/EdgeBinding'
+            type: array
+            items:
+              $ref: '#/components/schemas/EdgeBinding'
       required:
         - node_bindings
         - edge_bindings
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -223,14 +223,14 @@ components:
         answer the question. This graph is a representation of a question.
       properties:
         nodes:
-          type: array
+          type: object
           description: List of nodes in the QueryGraph
-          items:
+          additionalProperties:
             $ref: '#/components/schemas/QNode'
         edges:
-          type: array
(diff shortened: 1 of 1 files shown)


patrickkwang:
Yeah, there's definitely a case to be made that the ids should be duplicated in the values for expediency. It's the difference between
```{'id': node_id, **node} for node_id, node in nodes.items()```
and
```node for node in nodes.values()```

**Comment by edeutsch** [comment]
The benefit is that a node knows its own id. I foresee headaches if it doesn't.
The complication is that a node's internal id could be different than its nodes{} key (but shouldn't be).
I vote for nodes and qnodes et al. internally knowing their own ids.
Others votes?

**Comment by dkoslicki** [comment]
> The benefit is that a node knows its own id. I foresee headaches if it doesn't.
> The complication is that a node's internal id could be different than its nodes{} key (but shouldn't be).
> I vote for nodes and qnodes et al. internally knowing their own ids.
> Others votes?

Echoing this as well: all nodes and edges need to know their own ids.

A concern I have (that might be due to my not understanding the proposal fully) is the following:
When you @patrickkwang say that “The qg_id field of each becomes the associated key” are you referring only to the “results”? Not query graph (qg) or knowledge graph (kg) nodes/edges, right? I find it vitally important that qg and especially kg entities know their own qg_id’s.

**Comment by patrickkwang** [comment]
@dkoslicki, you understand right. We can do this for the bindings, the QG nodes/edge, and the KG nodes/edges. Just to clarify, if we do not include the ids in the QG and KG values, no information is being lost. If one wanted to, it is straightforward to convert the new map form into the old list form by doing something like
```
query_graph['nodes'] = [
    {'id': node_id, **node}
    for node_id, node in query_graph['nodes'].items()
]
```

The question is whether any inconvenience outweighs 1) the savings in payload size and 2) the risk of generating inconsistencies, as @edeutsch mentioned. In practice, there should rarely be a need to do the above transformation. Most of the time, I find myself either indexing into this map, in which case I already have the `id`, or iterating through it, in which case the use of `items()` is no more painful than  `values()`.

**Comment by edeutsch** [comment]
I'm also puzzled about a very trivial detail. In here:
image

If the new type is an object, then why is the new type under additionalProperties an array?

**Comment by patrickkwang** [comment]
These must be arrays to allow for multiple bindings to each query_graph node/edge. This is permitted in the current format just by adding multiple independent NodeBinding objects with the same qg_id.

**Comment by edeutsch** [comment]
But the kg_id inside a NodeBinding is already a list. So if you have multiple nodes for a qg_id, I thought you would do it that way? The proposal is to allow either way?
      "node_bindings": { "n01": [
         {
          "kg_id": [ "UniProtKB:Q16790", "UniProtKB:Q16791" ]
          "qg_id": "n01"
         },
       ]
     }

OR:

      "node_bindings": { "n01": [
         {
          "kg_id": [ "UniProtKB:Q16790" ]
          "qg_id": "n01"
         },
         {
          "kg_id": [ "UniProtKB:Q16791" ]
          "qg_id": "n01"
         },
       ]
     }

Is that what is means? I just don't quite understand the notation

**Comment by patrickkwang** [comment]
Ahhh, I see. You're right, there are currently two ways to do multiple bindings; this proposal does not fix that. But probably we should. I'm inclined to disallow the `kg_id`-list, because breaking them out into multiple objects allows unique metadata (provenance/evidence/score) for each binding. Thoughts?

Also note that the proposal includes removing the `qg-id` field, e.g.
```
"node_bindings": { "n01": [
     {
      "kg_id": "UniProtKB:Q16790",
      "provenance": {...}
     },
     {
      "kg_id": "UniProtKB:Q16791",
      "provenance": {...}
     },
   ]
 }
```

**Comment by edeutsch** [comment]
yes, this seems pretty good. There is of course no provenance field in the PR. Is that something you're already using? OR thinking might be useful someday? Seems like most provenance is in the KG rather than in the binding. What is the use case for providing provenance in the binding?

**Comment by patrickkwang** [comment]
No, the "provenance" field is just an example for now. We will definitely want provenance on the KG nodes/edges as well. In ROBOKOP and ARAGORN, we add some metadata to individual bindings to represent information like "how relevant is this node/edge _in this question_".

**Comment by edeutsch** [comment]
the latter there makes a lot of sense. Can you share how you encode that relevance in bindings?

**Comment by patrickkwang** [comment]
Under the current scheme, we do e.g.
```
{
  "node_bindings": [
    {
      "qg_id": "n00",
      "kg_id": "MONDO:0100096",
      "score": 0.72
    }
  ]
}
```

edeutsch:
that seems quite sensible. I've made a note in the Google doc about it for discussion.

oneilsh:
> I generally agree with the cleanliness of ids as map keys and not in contained objects, and it should be an easy refactor for most code bases, and it being before 1.0 is a good time for a breaking change.

That said, this also seems like a good time to plan for potential future breaking changes; I'm not sure if anything needs to be added to the API though to do so (like a "deprecations_notice" field). I've mostly been a consumer of deprecation in APIs, not a producer ;)

PR #106, opened 2020-06-16
Add 'description' property to Node

patrickkwang:
_on behalf of @edeutsch_


patrickkwang:
I believe that any field that is useful exclusively for humans should not be included in APIs that are not intended to be user-facing. It could be handy to have a utility that takes CURIEs and returns descriptions, which could be used by a GUI application. But if we send all that data over the Reasoner API, we waste bandwidth.

The same principle theoretically applies to the Node `name` as well. That one is less concerning because it will be smaller and less likely to be ignored or discarded later.

patrickkwang:
Closing on behalf of @edeutsch.

PR #107, opened 2020-06-16
add 'symbol' property to Node

patrickkwang:
_on behalf of @edeutsch_


vdancik:
I don't think symbol should be a specific node property as it is only applicable to genes. If there is a need for very short strings to be associated with nodes, we can consider to have 'abbreviation' or 'label' property.

patrickkwang:
I agree that properties that are not relevant to all node types should not be included like this. @edeutsch, would you like to make an argument in support of this addition?

patrickkwang:
Closing on behalf of @edeutsch.

PR #109, opened 2020-06-16
use 'node_attributes' in place of additionalProperties

patrickkwang:
_on behalf of @edeutsch_

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -260,11 +260,11 @@ components:
         type:
           type: string
           description: CURIE of the semantic type of the attribute, from the EDAM ontology if possible.
-          example: 'EDAM:data_1187'
+          example: EDAM:data_1187
         url:
           type: string
           description: Human-consumable URL to link out and read about the NodeAttribute (not the Node).
-          example: 'https://pubmed.ncbi.nlm.nih.gov/32529952'
+          example: https://pubmed.ncbi.nlm.nih.gov/32529952
         source:
           type: string
           description: Source of the NodeAttribute, as a CURIE prefix.
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -269,6 +269,9 @@ components:
           type: string
           description: Source of the NodeAttribute, as a CURIE prefix.
           example: UniProtKB
+      required:
+        - type
+        - value
       additionalProperties: false
     Edge:
       type: object
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -252,18 +252,25 @@ components:
       properties:
         name:
           type: string
-          description: Human-readable name or label for the NodeAttribute. Should be the name of the semantic type term.
+          description: >-
+            Human-readable name or label for the NodeAttribute. Should be the
+            name of the semantic type term.
           example: PubMed Identifier
         value:
           example: 32529952
-          description: Value of the attribute. May be any data type, including a list.
+          description: >- 
+            Value of the attribute. May be any data type, including a list.
         type:
           type: string
-          description: CURIE of the semantic type of the attribute, from the EDAM ontology if possible.
+          description: >-
+            CURIE of the semantic type of the attribute, from the EDAM ontology
+            if possible.
           example: EDAM:data_1187
         url:
           type: string
-          description: Human-consumable URL to link out and read about the NodeAttribute (not the Node).
+          description: >-
+            Human-consumable URL to link out and read about the NodeAttribute
+            (not the Node).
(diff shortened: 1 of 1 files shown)


vdancik:
I think this should be in core as node attributes are more suitable for machine reasoning then additional properties.

rtroper:
I agree with Vlado, I think this should be in core. It seems structured enough for reasoners to make use of. If it does prove useful in this respect, this change may be a good bridge to further standardization (later) for full-fledged automated reasoning.

brettasmi:
Per last week's discussion and a review of the changes, I think this is the way to go 👍

patrickkwang:
I think that instead of putting these extra attributes and their associated metadata in the response payload, the keys/values should be included directly, as allowed by `additionalProperties=true`, and the associated metadata should be included in the server's individual OpenAPI schema. For example, rather than sending
```
{
  'id': 'HGNC:7898,
  'node_attributes': [
    {
      'name': 'gene_symbol',
      'name': 'NPC1',
      'type': 'EDAM:data_1026'
    }
  ]
}
```
One can send
```
{
  'id': 'HGNC:7898',
  'gene_symbol': 'NPC1'
}
```
and include the following in the service's `Node` schema properties:
```
gene_symbol:
  type: string
  format: 'EDAM:data_1026'
```

CaseyTa:
@patrickkwang I thought part of the reasoning for having NodeAttributes was to standardize how we communicate various items across services? Wouldn't we lose that standardization if each service can define their own schema? Sorry if I'm misunderstanding something. 

> 
> 
> I think that instead of putting these extra attributes and their associated metadata in the response payload, the keys/values should be included directly, as allowed by `additionalProperties=true`, and the associated metadata should be included in the server's individual OpenAPI schema. For example, rather than sending
> 
> ```
> {
>   'id': 'HGNC:7898,
>   'node_attributes': [
>     {
>       'name': 'gene_symbol',
>       'name': 'NPC1',
>       'type': 'EDAM:data_1026'
>     }
>   ]
> }
> ```
> 
> one can send
> 
> ```
> {
>   'id': 'HGNC:7898',
>   'gene_symbol': 'NPC1'
> }
> ```
> 
> and include the following in the service's `Node` schema properties:
> 
> ```
> gene_symbol:
>   type: string
>   format: 'EDAM:data_1026'
> ```
> 
> This is at least equally expressive, and keeps payloads small.

CaseyTa:
What is the interpretation of multiple entities in the NodeAttributes list? Does each come from a different source? Is the list ordered by preference or unordered? Should each entry in the list fill as many fields in NodeAttributes as possible?

webyrd:
I favor 'node_attributes' over additionalProperties.

@CaseyTa @patrickkwang I am also curious about Patrick's proposal to have different schema.  How exactly would this work?

patrickkwang:
Re: @CaseyTa
> I thought part of the reasoning for having NodeAttributes was to standardize how we communicate various items across services?

Yes, there are some items that are necessary and common to all services and should be standardized, but not everything. NodeAttributes and additionalProperties serve the same purpose: to allow inclusion of those non-standardized items.

> Wouldn't we lose that standardization if each service can define their own schema?

No. Using either method, the client should be able to safely ignore the extra stuff and still get what they need. But there is an important and subtle issue here that @webyrd identified:

> I am also curious about Patrick's proposal to have different schema. How exactly would this work?

If each server's OpenAPI schema is slightly different, we need a concept of API subclassing, for lack of a better term. This is relatively easy to understand for the JSON responses themselves. The object returned by a service is expected to conform to the service's individual schema AND to the standard Message schema. If we allow additionalProperties in the standard, then services can define extra properties of their own at will, as long as they do not _require_ them, and meet this condition. If you look to apply this "subclassing" idea more broadly, it breaks down a little. An OpenAPI schema is not designed to do what we're doing with it. It defines a single specific API, rather than constraining a set of APIs. But we've danced around this up to now, and I think we can continue to do so if the variations are mostly in the request/response body.

patrickkwang:
Two small modification have been made (https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI/pull/109/commits/92759277e0f21318d1b32adced00a7cdb05d4272 and https://github.com/NCATS-Tangerine/NCATS-ReasonerStdAPI/pull/109/commits/9908212a043d14d6146ba8faa28d25c1bf06312e). As a result, all reviews have been dismissed as stale.

Please re-review.

patrickkwang:
I am for this change as a whole because it has the following benefits:
1. provides a standardized way to encode attribute metadata via the `type` property
2. allows multiple attributes with the same name/type (which a dict could not do)
3. allows pass-through of attribute metadata, e.g. KP -> ARA -> ARS (which could not be accomplished via OpenAPI schema)

CaseyTa:
I prefer node attributes over using additional properties.

webyrd:
Looks good to me.

PR #110, opened 2020-06-16
tweak Node description

patrickkwang:
_on behalf of @edeutsch_


patrickkwang:
This a "minor documentation change" that will be accepted after a week provided only that no one objects.

patrickkwang:
👍

I think this is a good change, though apparently GitHub will not allow me to Approve because I technically opened the PR.

PR #111, opened 2020-06-30
Add EdgeAttribute and 2nd-class attributes to Edge

edeutsch:
This pull request adds the EdgeAttribute component and also a large set of "second-class" attributes that we've long said are desirable for each edge and were based on the Knowledge Graph Standardization effort from long ago.

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -243,22 +243,89 @@ components:
       additionalProperties: true
     Edge:
       type: object
-      description: An edge in the thought subgraph linking two nodes
+      description: An edge in the knowledge graph linking two nodes
       properties:
         type:
           $ref: '#/components/schemas/BiolinkRelation'
         source_id:
           type: string
-          example: 'https://omim.org/entry/603903'
-          description: Corresponds to the @id of source node of this edge
+          example: 'OMIM:603903'
+          description: Corresponds to the map key CURIE of the source node of this edge
         target_id:
           type: string
-          example: 'https://www.uniprot.org/uniprot/P00738'
-          description: Corresponds to the @id of target node of this edge
-      additionalProperties: true
+          example: 'UniProtKB:P00738'
+          description: Corresponds to the map key CURIE of the target node of this edge
+        edge_attributes:
+          type: "array"
+          description: "A list of additional attributes for this edge"
+          items:
+            $ref: "#/components/EdgeAttribute"
+        negated:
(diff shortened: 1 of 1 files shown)


edeutsch:
Can we get these provenance questions in front of the data modeling group soon? These were decisions that were made long ago, and that we (ARAX) implemented. If these will change, that's fine, but I'd like to be sure that that group is actively working on this? Who can put these in front of that group and make sure it high on their priority list?

confidence=0 does *not* imply negated=true. confidence=0 means you have zero confidence in the assertion that is being made, be it negated or not. One should probably not be making assertions that have zero confidence, but that's how it would be expressed. The assertion could be that A does NOT interact with B with confidence 0.80. Is this the same as A interacts with B with confidence 0.20? Maybe.

CaseyTa:
> I think we should accept a version of #112 and perhaps break the other properties into separate PRs, if there is interest in them.

I agree with @patrickkwang on this. Would make it easier to get the changes in, and I view these two sets of changes as independent. 

For whatever it's worth, I don't think confidence=0 implies negated=true. I interpret it as meaning that the source does not know whether or not the edge relationship is true, not that it knows that the opposite of the relationship is true. But on that note, I'm not sure if negated on a relationship like "upregulates" means 1) that there is no relationship between the two nodes; 2) if the source node downregulates the target node; or 3) either #1 or #2 are true. Once these definitions are clarified, it would be helpful to have them well documented. 

Is ```relation``` used when the source knows the relationship to a finer granularity than Biolink Model allows? I assume ``type``` is restricted to Biolink Model.

patrickkwang:
I didn't mean to start a debate here, but just to illustrate that there are nontrivial questions to answer around confidence/negation. I have requested that the data modeling committee prioritize that issue. They are already engaged in discussions around provenance.

edeutsch:
> Is relation used when the source knows the relationship to a finer granularity than Biolink Model allows? I assume type is restricted to Biolink Model.

yes, exactly!

ehinderer:
I agree  with @patrickkwang that these are provenance issues that are being worked out with the data modeling group presently. Not sure if I have anything of value to recommend for the time being, but if you're interested in that provenance  discussion it's going on here: https://github.com/biolink/biolink-model/issues/357. 

I can suggest this for an agenda item for the next data modeling meeting.

edeutsch:
This PR is declined for now. These items remain in `extended` but any potential additions to the core are deferred until the data modeling group decide how to model provenance information.

PR #112, opened 2020-06-30
Add EdgeAttribute only

edeutsch:
This pull request provides a minimalist addition of only EdgeAttribute (and removes additionalProperties: true) such that all attributes that we want to have on our edges all go in EdgeAttributes. This is alternative to the other pull request with multiple "second-class" attributes. One PR or the other, or somewhere in between?

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -290,17 +290,17 @@ components:
           $ref: '#/components/schemas/BiolinkRelation'
         source_id:
           type: string
-          example: 'OMIM:603903'
+          example: OMIM:603903
           description: Corresponds to the map key CURIE of the source node of this edge
         target_id:
           type: string
-          example: 'UniProtKB:P00738'
+          example: UniProtKB:P00738
           description: Corresponds to the map key CURIE of the target node of this edge
         attributes:
-          type: "array"
-          description: "A list of additional attributes for this edge"
+          type: array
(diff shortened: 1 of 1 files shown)


API/TranslatorReasonersAPI.yaml
```diff
@@ -243,22 +243,54 @@ components:
...
+          example: 'OMIM:603903'
+          description: Corresponds to the map key CURIE of the source node of this edge
         target_id:
           type: string
-          example: 'https://www.uniprot.org/uniprot/P00738'
-          description: Corresponds to the @id of target node of this edge
-      additionalProperties: true
+          example: 'UniProtKB:P00738'
+          description: Corresponds to the map key CURIE of the target node of this edge
+        edge_attributes:
```
patrickkwang, line 258:
```suggestion
        attributes:
```

patrickkwang:
There are some styling (cosmetic) inconsistencies here that we should establish conventions for. Specifically, where should we use quotes for string values, and what kind? We could use something like [this](https://yamllint.readthedocs.io/en/stable/) to enforce style.

patrickkwang:
@bill-baumgartner, Yes. I agree they ought to be unified. Perhaps we can rework this PR after #109 is merged.

patrickkwang:
@bill-baumgartner Thanks for catching that! It's fixed now.

webyrd:
Looks fine to me.

PR #114, opened 2020-07-02
Added "translator" and "reasoner" tags

newgene:
Adding "translator" and "reasoner" tags in this template to indicate an API as a "Translator's KP API implemented as ReasonerStdAPI", when [registered in SmartAPI registry](https://smart-api.info/add_api).

"translator" tag is included for all Translator KP APIs, and "reasoner" tag is specific for Reasoner APIs.

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -25,7 +25,7 @@ tags:
       description: Documentation for the reasoner query function
       url: http://reasonerhost.ncats.io/overview.html#query
   - name: translator
-  - name: reasoner 
+  - name: reasoner
 paths:
   /predicates:
     get:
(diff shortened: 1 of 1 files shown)


ehinderer:
Is the "reasoner" name outdated? I thought that "reasoner" was the old name for what we're calling an ARA now.

newgene:
@edeutsch @ehinderer it does not have to be called "reasoner", can be other name after the rebranding. But I think it's necessary to add a tag to indicate the implement type of the KP APIs, "reasoner" KP API is one of them.

patrickkwang:
@newgene Should these tags be associated with specific operations? My reading of the OpenAPI Specification is that this top-level tag list is meant just to provide metadata around the tags used by specific operations. Listing the reasoner tag as you've done should not be taken to imply that every endpoint described by the OpenAPI schema is a Reasoner API endpoint, as these services may provide arbitrary additional operations. Requiring the reasoner tag on the /query endpoint seems sensible, though.

vdancik:
We can call it Knowledge-Exchange API (KEXAPI)

webyrd:
> We can call it Knowledge-Exchange API (KEXAPI)

Not bad!  :)

edeutsch:
Sentiment seems to be strongly in favor of approval with a proviso that we may alter it again later based on the decisions of the renaming. I'll hold it open until Monday and then merge unless there is opposition.

PR #116, opened 2020-07-08
Create Policies.md

edeutsch:


edeutsch:
We have several approvals. We'll leave the floor open until Monday and then merge it unless there is oppo.

patrickkwang:
Looks good. One minor question. Why is the Travis badge on this doc?

edeutsch:
The reason is only that it was in the template that I used/example that I followed. Should it be in all documents? Or just in the top-level README? I'm happy to remove it if it doesn't seem appropriate.

Policies.md
```diff
@@ -0,0 +1,18 @@
...
+
+This document described development and release policies for the NCATS Translator API
+
+## Repository branches
+- The repository shall have two primary branches, `master` and `experimental`
+- The `master` branch shall contain the current core API schema, both in-development changes and tagged releases
+- The `experimental` branch shall contain the current extended API schema, a definition of additional properties that are used and promoted by at least one team
+
+## Version Numbering
+- After a tagged release, the very next commit to the `master` branch will receive an increment to the next minor version number with 'dev' appended
```
patrickkwang, line 11:
```suggestion
- After a tagged release, the very next commit to the `master` branch will receive an increment to the next minor version number with '-dev' appended
```
per https://semver.org/#spec-item-9

Policies.md
```diff
@@ -0,0 +1,18 @@
...
+  (e.g. 0.9.3dev)
```
patrickkwang, line 12:
```suggestion
  (e.g. 0.9.3-dev)
```
per https://semver.org/#spec-item-9

bill-baumgartner:
Should there be a policy/procedure included for promoting "experimental" features to the main branch? Should a review be prompted after a certain period of time, or can features live in `experimental` in perpetuity?

edeutsch:
We seem to be headed for a path where only the very minimal items will live in the core and many features will live in `experimental` in perpetuity. I rather prefer the name `extended` since that seems to capture the concept better.

bill-baumgartner:
> We seem to be headed for a path where only the very minimal items will live in the core and many features will live in `experimental` in perpetuity. I rather prefer the name `extended` since that seems to capture the concept better.

In that case, I completely agree with you that `extended` would be a better name for the `experimental` branch.

PR #117, opened 2020-07-09
add QNode.is_set

edeutsch:


edeutsch:
This PR is currently in limbo while @patrickkwang checks with @cmungall about how to model this in BioLink: combination (is_set)+(protein) or (protein_set).

ehinderer:
I believe this issue is related to this https://github.com/biolink/biolink-model/issues/385

edeutsch:
Exactly! Thanks for the link, I couldn't find it. I suppose we wait here until there is some resolution there.

ehinderer:
Are we waiting for a response from data modeling on this?

edeutsch:
In re-examining the pull request, I realize that we can provide a better definition now after our discussion. So before you approve this, how about we change the description to:

Boolean that if set to true, indicates that this QNode MAY have multiple KnowledgeGraph Nodes bound to it within each Result, implying that those Nodes should be considered as an independent set, rather than individually.

and I'll add a "default: false" attribute and remove the example.

comments?

CaseyTa:
How about:

> Boolean that if set to true, indicates that this QNode MAY have multiple KnowledgeGraph Nodes bound to it within each Result. The nodes in a set should be considered as a set of independent nodes, rather than a set of dependent nodes, i.e., the answer would still be valid if the nodes in the set were instead returned individually. 

Is more than one QNode allowed to have is_set=True? If so, how should that be interpreted? Would that mean that all possible combinations of nodes across all sets should each be a valid answer?

patrickkwang:
Yes, I think multiple nodes should be allowed to have `is_set=true`. But they do not need to be _densely_ connected. Imagine a query graph:
```
(n0)-[e01]-(n1, is_set=true)-[e12]-(n2, is_set=true)-[e23]-(n3)
```
In each result, each n1 must be connected to n0 and to at least one n2. Each n2 must be connected to n3 and at least one n1.

CaseyTa:
Suggested additional details in the description based on Patrick's comment:

> Boolean that if set to true, indicates that this QNode MAY have multiple KnowledgeGraph Nodes bound to it within each Result. The nodes in a set should be considered as a set of independent nodes, rather than a set of dependent nodes, i.e., the answer would still be valid if the nodes in the set were instead returned individually.
>
> Multiple QNodes may have is_set=True. If a QNode (n1) with is_set=True is connected to a QNode (n2) with is_set=False, each n1 must be connected to n2.  If a QNode (n1) with is_set=True is connected to a QNode (n2) with is_set=True, each n1 must be connected to at least one n2.

edeutsch:
Underlying file has moved to a new location and previous approvals are invalid, so let's just close in favor of a fresh PR #149

PR #118, opened 2020-07-09
add a mechanism to provide logging information

edeutsch:
I am trying to propose this to the `experimental` branch.

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -434,6 +434,11 @@ components:
           description: >-
             Dict of options that can be sent with the query. Options are tool
             specific and not stipulated here
+        log:
+          description: "List of informational 'logging' text lines from the server about how the server went about satisfying the query. e.g. logged INFO, WARNING, and ERROR message lines"
+          type: array
+          items:
+            type: string
         results:
           description: List of all returned potential answers for the query posed
           type: array


webyrd:
I'm trying to understand the intent of this logging information.  Is the purpose to be able to diagnose a mis-configured web server or ARA or KP?  As opposed to tracking the reasoning involved, or data provenance, for example?  Or is the intent to capture and report semantic or reasoning information, in addition to standard server-related errors?

vdancik:
Logging sounds reasonable for development branch only

edeutsch:
FYI, here is an example of how we are using log currently:
https://arax.rtx.ai/beta/?m=2650
- Click on [Messages] on the left to see a view of the logging that is returned.
- if you want to see the JSON, click on [Dev Info] and browser-search for `"log":` including the double quotes
I'm not currently proposing a specific structure within `log`, but just that there could be one and that's how we're doing it.

edeutsch:
In response to @webyrd questions:
>I'm trying to understand the intent of this logging information. Is the purpose to be able to diagnose a mis-configured web server or ARA or KP? 

Probably not the web server, but the ARA or KP yes. More likely the ARA than a KP.

>As opposed to tracking the reasoning involved, or data provenance, for example? Or is the intent to capture and report semantic or reasoning information, in addition to standard server-related errors?

It is intended to convey any sort of free text information about the process of reasoning or answering the query that a user might like to know. (or in the case of WARNINGS or ERRORs) probably needs to know. In the above example, there is a warning because a set of proteins was provided but is_set was set to false. User intent was clear, so instead of erroring out, we just issue a warning. Also in the set of proteins, one is not recognized. Again, we could error out, but we decided to just issue a WARNING and proceed with the rest of the list. The user can then decide whether to be content with what was done or issue an adjusted query. And then there's lots of INFO information. Probably too much. It gives some insight into what is going on behind the curtain. At the moment, we like it for testing and debugging, but see its value for consumers.

edeutsch:
Maybe this will be expanded in the future, but for now, let's get this into `extended` as a top-level container as is.

PR #119, opened 2020-07-16
Add document description metadata for Message

edeutsch:
These Message metadata fields are currently in the experimental branch. Do they deserve to be in the core schema?
At issue really is: do we want to encourage each returned Message to be a standalone document that can be identified and be self-describing?

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -138,7 +138,7 @@ components:
           type: string
           example: '2018-01-09 12:34:45'
           description: Datetime string for the time that this message was generated
-          additionalProperties: true
+      additionalProperties: true
     Result:
       type: object
       description: One of potentially several results or answers for a query
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -108,7 +108,37 @@ components:
             referenced in any of the possible answers to the query OR
             connection information for a remote knowledge graph
           $ref: '#/components/schemas/KnowledgeGraph'
-      additionalProperties: true
+        context:
+          type: string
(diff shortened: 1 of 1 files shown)


webyrd:
I'm not sure I understand the 'JSON-LD context URI' part of the message document description.  What is this for?

ehinderer:
I also do not understand the 'JSON-LD context URI' but I do think that the metadata described in these changes do seem deserving for the core schema. I believe this would help with Vlado's concern about tracing from which KPs answers originate.

edeutsch:
I have updated the content URI to a real example:
https://raw.githubusercontent.com/biolink/biolink-model/master/context.jsonld
This might be considered a Translator-wide one.
But we (ARAX) have CURIEs that are not in BioLink, so we would have a local ARAX-wide context.jsonld that would define our curies properly.

edeutsch:
What we were aiming toward when this was originally designed was JSON-LD. A public example of JSON-LD (https://json-ld.org/playground/) goes like this:
```
{
  "@context": "http://schema.org/",
  "@type": "Person",
  "name": "Jane Doe",
  "jobTitle": "Professor",
  "telephone": "(425) 123-4567",
  "url": "http://www.janedoe.com"
}
```
where @context, @type, @id seem like classic JSON-LD attributes that are commonly provided in self-describing linked data. The @ symbols seemed irritating so we opted to dispense with that character and simply provide context, type, and id.

Another example: https://github.com/prefixcommons/biocontext/blob/master/examples/hpoa_example.jsonld

While I agree that the context is probably global to the endpoint, it seems customary in JSON-LD to explictly provide it. I know shockingly little about JSON-LD but was led to believe we were fumbling in the correct direction toward useful semantic web documents by including these properties.

edeutsch:
Decided to leave all this all in `extended` and consider individual items at some point in the future.

PR #120, opened 2020-07-16
Remove all Feedback components and Result id

edeutsch:


webyrd:
Not sure I understand this change.  Where is the feedback coming from?  KPs?  ARAs?  End user?  Is there another way to capture this feedback, if it might be useful?

ehinderer:
> Not sure I understand this change. Where is the feedback coming from? KPs? ARAs? End user? Is there another way to capture this feedback, if it might be useful?

Same question from me...

edeutsch:
> Not sure I understand this change. Where is the feedback coming from? KPs? ARAs? End user? Is there another way to capture this feedback, if it might be useful?

The recommended change is to *remove* from `experimental` the "Feedback" part of the schema. This was proposed by ARAX and implemented by ARAX but has not seen much use and as we move to more of an ARS->ARA model, these seems even less useful. In the spirit of cutting away chaff that is not really used, I propose to get rid of it.

Or original implementation (done during the feasibility phase) is summarized as follows:
- Each Message and Result within a Message gets a unique id
- End users (humans) who view our Messages and Results have the opportunity to provide "feedback": a graded rating and free text comments to answers.
- In anyone subsequently views that earlier result or asks the same exact question, they see not just the response, but also the response decorated with previous human feedback.
- Feedback was managed by our Javascript GUI, but all input/output went through this API so any client could read/write feedback
- We had fantasies of improving reasoning based on feedback.
- In reality we got hardly any. And if the ARS is now going to be our main client, there is even less opportunity to gather human feedback. No one else implemented it to my knowledge.
- So I'm prepared to tombstone it. It can always be exhumed from GitHub in the future when its time comes.
- Of course, if another group is eager to implemented it, too, we can reconsider

ehinderer:
I think we'll get more feedback as the project matures and gets more use. I was under the impression that community feedback/engagement was a major factor in the Translator initiative so I hesitate to approve of a change that merely gets rid of work done toward that end. 

Besides, some ARAs--for instance the Explanatory Agent--will eventually rely on some level of feedback to better inform the quality of natural language understanding algorithms and fine-tune machine learning methods. Part of our research plan includes incorporating community feedback, and we have team members specifically for that job.

If we want to **ignore** this for the time being, though, I think that's perfectly reasonable.

edeutsch:
General consensus is that this prior work may yet be useful to several groups, so do not remove at this time. Leave as is and groups interested in implementing Feedback should look to this as a starting point and initiate a discussion when they are planning their own implementation.

PR #123, opened 2020-07-23
Restore components inadvertantly deleted in 0.9.2

edeutsch:
These components were present in 0.9.1, but were accidentally deleted in the transition to a core/experimental split in 0.9.2. This PR restores these components to `experimental` branch only.


edeutsch:
Decided to withdraw in favor of proposing things piecemeal.

PR #125, opened 2020-07-28
Re-add Query.bypass_cache to extended

edeutsch:
The Query class used to have a bypass_cache attribute, but it was accidentally deleted in 0.9.1 -> 0.9.2
Use this flag to bypass potential answer caching


patrickkwang:
Closing in favor of #139

edeutsch:
We decided to implement this as a query (URL) parameter rather than in the payload.

PR #126, opened 2020-07-28
Restore Query.asynchronous to extended

edeutsch:
The Query class used to have an asynchronous attribute, but it was accidentally deleted in 0.9.1 -> 0.9.2
Use this flag to signify alternate behavior than the default 'wait for the message' behavior.

Set to 'true' in order to receive an incomplete message_id if the query will take a long time. The client can then periodically poll with that message_id for a status update and to receive an eventual complete message.

Set to 'stream' in order to receive newline-delimited logging messages describing progress in answering the query, with the last line being the resulting message in its entirety.

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -270,6 +270,16 @@ components:
       properties:
         message:
           $ref: '#/components/schemas/Message'
+        asynchronous:
+          type: string
+          example: 'true'
+          description: >-
+            Set to 'true' in order to receive an incomplete message_id if the query
+            will take a long time. The client can then periodically poll with that
+            message_id for a status update and to receive an eventual complete message.
+            Set to 'stream' in order to receive newline-delimited logging messages
+            describing progress in answering the query, with the last line being
+            the resulting message in its entirety.
       additionalProperties: true
       required:
         - message


patrickkwang:
Let's resolve #130 first.

vdancik:
Would this mean that anyone implementing extended API must support asynchronous communication? I would prefer to have a different endpoints, like /assyn_query and /paged_query rather than having this in the payload.

edeutsch:
> Would this mean that anyone implementing extended API must support asynchronous communication

I would say no. The extended API is not intended as an all-or-nothing set, but rather a shared recommendation on how to implement certain ideas. *If* a server chooses to implement asynchronous communication, let's all implement it in the same way rather than 5 different ways. It might be nice if servers that did not fully support it would return an error if they received this option but I don't feel strongly about it.

>  I would prefer to have a different endpoints, like /assyn_query and /paged_query rather than having this in the payload

This is another way, but what if we have several options and some are combinable? /assyn_coreonly_query, /assyn_extended_query? It seems far easier on clients and servers to have one endpoint with settable switches. But maybe I'm just lazy.

vdancik:
I assume that asynchronous query returns a totally different JSON than synchronous query thus warranting a different endpoint. Also asynchronous query needs another endpoint to retrieve response.

While core/extended business might also be implemented via endpoints, I assume that every server will be either core or extended and we can indicate that when we register ARAs or KPs, thus client knows whether to submit core or extended query. If client wants core response from extended server, it can indicate possibly via query parameter as suggested in #130.

endpoints in summary:
|core/extended|synchronous|endpoints|
|-----|--------------|-----------|
|core|synchronous|`/query`|
|core|asynchronous|`/query/submit` and`/results`|
|extended|synchronous|`/query?core=boolean` with `false` default|
|extended|asynchronous|`/query/submit?core=boolean` and`/results`|

edeutsch:
We decided to implement this differently with multiple new endpoints. Patrick will create a PR with a different design.

PR #130, opened 2020-07-28
Add Query.use_core_schema property

edeutsch:
This adds an explict property use_core_schema to Query.
Set to true in order to receive a Message that is limited only to named properties in the core schema. Properties from the `extended` schema and other additional properties must not be transmitted.

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -85,6 +85,13 @@ components:
       properties:
         message:
           $ref: '#/components/schemas/Message'
+        use_core_schema:
+          type: string
+          example: 'true'
+          description: >-
+            Set to true in order to receive a Message that is limited only to named properties
+            in the core schema. Properties from the `extended` schema and other additional
+            properties must not be transmitted.
       additionalProperties: true
       required:
         - message


patrickkwang:
This seems to imply that the core and extended APIs are to be treated as incompatible alternatives.

How is the extended API (going to be) used by servers and clients?

servers implementing the Reasoner API
---
core only
  * returns responses compliant with the core schema, but not necessarily the extended schema
  * this may include additional properties where permitted by `additionalProperties != false`
  * these additional properties may violate the prescriptions of the extended standard, e.g. `tool_version: 2` (this is supposed to be type string)
  * query options from the extended standard (like the proposed `asynchronous`) may be ignored

extended
  * returns responses compliant with the extended schema
  * this may or may not make use of the extended properties (none of which are `required`)
  * implements all query options from the extended standard

clients calling Reasoner-API servers
---
core only
  * sends requests compliant with the core schema, but not necessarily the extended schema
  * that this may include additional properties where permitted by `additionalProperties != false`
  * these additional properties may violate the prescriptions of the extended standard, e.g. `QEdge.negated: 0` (this is supposed to be type boolean)
  * ignores extended properties in responses, e.g. `Edge.negated`

extended
  * sends requests compliant with the extended standard
  * this may or may not make use of the extended properties (none of which are `required`)

---

It seems to me that the core and extended APIs _are_ currently incompatible in that clients and servers that do not match will not communicate effectively. This incompatibility is driven by a few extended properties that change the interpretation of a query or result. Specifically:
1. `Edge.qualifiers`
2. `Edge.negated`
3. `QEdge.negated`
4. `Message.query_options`, potentially
as well as the proposed `Query.asynchronous` and `Query.PMPP` (I think). There may be more.

`Query.use_core_schema` is one way to resolve this incompatibility, by effectively ratifying the bifurcation of this standard. I would rather remove the problematic properties or promote them to the core schema. Then the extended standard would serve as a truly optional definition of additional features that one can implement/support as desired, and retain compatibility with servers/clients implementing only the core standard.

edeutsch:
> This seems to imply that the core and extended APIs are to be treated as incompatible alternatives.

I did not intend to imply that. The use case I'm thinking of is the ARS. I was under the impression that you wanted the core schema as lean as possible because the ARS wanted the leanest response with no additional information. So I propose this option so that the ARS can tell an ARA to strip down the response Message so that only core elements remain. Our ARAX process enriches the Message with lots of information, but I envision adding a routine in our code that will strip out everything but the core elements if this option is set. And leave in all the nice details if not set. Our GUI client and presumably most other clients will want a rich response, but we can strip it down to the core for the ARS, or any other client that wants a lean response.

patrickkwang:
If this is intended as a directive for servers implementing the extended schema to prune their responses, the PR should be directed at the extended branch. I'd also suggest renaming to something like `minimal_response`.

edeutsch:
It seems odd to me that a client should have to use the `extended` schema to tell the server *not* to use the `extended` schema. 

It seems to me that if we acknowledge that there are two worlds, the core and the extended then the switch should be in both worlds, core and extended. I only proposed it to core, with the assumption that all properties in core will be mirrored to extended when then are accepted.

I don't have a strong opinion on the name. Seems like if this is a directive, it should have a verb?

edeutsch:
Although I am somewhat against having different endpoints for multiple options as suggested in #126 by @vdancik, maybe I could come around to having a separate endpoint for extended, e.g. /query and /query_ext or something like that, where /query would support only core components and /query_ext would allow any combination of components from `extended`.

In our implementation, we would probably just call the same top level method with the extended=true/false, so little point to it IMO, but if that seemed preferable to others, I'd be fine with that.

vdancik:
I just suggested something similar/alternative in #126:

"While core/extended business might also be implemented via endpoints, I assume that every server will be either core or extended and we can indicate that when we register ARAs or KPs, thus client knows whether to submit core or extended query. If client wants core response from extended server, it can indicate possibly via query parameter:"

|core/extended|synchronous|endpoints|
|-----|--------------|-----------|
|core|synchronous|`/query`|
|core|asynchronous|`/query/submit` and`/results`|
|extended|synchronous|`/query?core=boolean` with `false` default|
|extended|asynchronous|`/query/submit?core=boolean` and`/results`|

vdancik:
We can also consider to add a metadata endpoint, say `\info` that would return information what server is and what it can do:

```
{
    "name": "best ARA ever",
    "version": "1.1.0",
    "synchronous support": true,
    "asynchronous support": false,
    "extended": true
}
```

(_maybe this should be an issue on it's own_)

edeutsch:
The consensus on the call is that trimming down the response is not a necessary function. Remove this PR. All servers can provide the information they want.

southalln:
I might add two additional comments. 

1) Vlado's principle above about the what goes into the message, but the how should be specified in the request path and not in the message was not fully decided, but positively received

2) There should be a general principle for this API that the inclusion of extended schema elements should not change the meaning of the message. If this is true, then ignoring any non-core schema elements (such as by the ARS) should not have deleterious consequences (it might reduce interpretability or confidence) - and the element requested here is not needed.

edeutsch:
Withdrawing..

Issue #115, opened 2020-07-02
Re-name NCATS-ReasonerStdAPI


patrickkwang:
We'd like to do this in concert with the move from the NCATS-Tangerine group to the NCATSTranslator group so that any technical juggling can happen just once.

We will also unify the naming across the repository, the `TranslatorReasonersAPI.yaml` file, and the documentation.

Please comment with suggestions!

ehinderer:
I heard someone suggest "TranslatorAPI." That person read my mind.

patrickkwang:
"Translator API" is clear and simple. My only concern is that it may not be specific enough. Are the SmartAPI-annotated, non-standardized KP APIs not Translator APIs? What about the interface that the ARS provides?

patrickkwang:
One option is just "Reasoner API" (in contrast with other variations "[NCATS/Translator] Reasoner [Std] API"). We'd have to accept that we're co-opting this term "reasoner" from any rational usage (including the tools from the feasibility phase) and that it's not descriptive in the least. But this has the advantage that everyone in Translator will already know what we're talking about.

patrickkwang:
Another option is something descriptive like "Query Graph API".

edeutsch:
> "Translator API" is clear and simple. My only concern is that it may not be specific enough. Are the SmartAPI-annotated, non-standardized KP APIs not Translator APIs? What about the interface that the ARS provides?

I like "Translator API".
It is not clear to me that "SmartAPI-annotated, non-standardized KP APIs" should be called "Translator APIs". I would call them "SmartAPIs" or "Translator SmartAPIs". As opposed to THE "Translator API".
The interface that the ARS provides would also use the "Translator API" I would say. I think we aim it to be the same?
I think "Translator API" is the best option I've seen so far. Feel free to disagree or suggest others.

andrewsu:
I like "Translator Graph API" and "Translator SmartAPI", because the graph nature is one distinguishing feature between the two approaches. (On the other hand, Translator Graph APIs could also be registered in the SmartAPI registry, so that is a potential confusion...)

vdancik:
We can call it Knowledge-Exchange API (KEXAPI)

PR #131, opened 2020-07-28
A first step toward updating the documentation

edeutsch:
Every journey begins with a single step.

### DIFF
--- a/API/README.md
+++ b/API/README.md
@@ -1,26 +1,28 @@
-# NCATS Translator Reasoners API
+# NCATS Translator Graph API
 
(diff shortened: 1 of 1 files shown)


API/README.md
```diff
@@ -1,27 +1,29 @@
-# NCATS Translator Reasoners API
+# NCATS Translator Graph API
```
patrickkwang, line 1: Looks like you have opinions about #115 😄.

edeutsch: ha ha, you caught me! Just checking to see if anyone is reading these. I don't actually have a string opinion. It seemed like Reasoners was out, so I thought I'd try what seemed like the best suggestion I've seen on for size.

edeutsch: I reverted the name change. How about everything else? Approvals for the rest?

PR #136, opened 2020-07-30
Use CURIEs as types

patrickkwang:


API/TranslatorReasonersAPI.yaml
```diff
@@ -312,16 +312,20 @@ components:
         - source_id
         - target_id
     BiolinkEntity:
-      description: A subclass of category named_thing (snake_case)
+      description: >-
+        CURIE for a Biolink class, NamedThing or a child thereof (PascalCase)
       type: string
+      pattern: 'biolink:[A-Z][a-zA-Z]*'
```
patrickkwang, line 318: Requesting feedback:
1. Do we want to specify that these must be Biolink CURIES? Or will any prefix do?
2. Do we want to specify that the "reference" after the colon must follow the Biolink conventions of PascalCase for classes and snake_case for slots?

balhoff: I would like to provide a specific `RO:` identifier as the predicate, and include one or more Biolink edge types as `type`—same as how nodes work.

edeutsch: I think it is nice to be explicit about what is wanted here, but I suppose we then need a policy about what to do when a biolink curie is not available. Suppose I have a type DefenseMechanism that I want to express and it's not in biolink. To be compliant, do I need to delete it? Can I lie and use biolink:DefenseMechanism even though it doesn't exist? Just translate these to biolink:NamedThing and just lose the information?

Maybe it is better not to be so strict in the schema and just build intelligence into the validator to spit out warnings when the value is not biolink:xxxxx so that a person could examine by hand and decide how to address..

patrickkwang: @balhoff The specific change proposed here affects the `type` field for nodes and edges. Are you proposing that `type` always come from Biolink, but that more precise identifiers can be provided in another field?

@edeutsch That is another potential solution to the issue you describe. We could have something like `subtype` for nodes that allows specification of a more granular non-Biolink class. That said, I think I agree that we should not enforce the `biolink` prefix here.

balhoff: @patrickkwang ah I see, I misinterpreted. I think think more precise identifiers for edges should be allowed (as in the `curie` for nodes). I think `type` should just come from Biolink. If more precise identifiers are provided in `curie` I guess I don't see the point of providing a non-Biolink `type` as well. Sorry if I am coming in out of left field here!

dkoslicki: >That said, I think I agree that we should not enforce the biolink prefix here.

Agreed, especially given @cmungall 's comment about it being acceptable to make up a CURIE prefix for something that is not (yet) in the BioLink model.

Also, to @edeutsch 's point, iirc, on the call just 40min ago, @cmungall said lying and using `biolink:DefenseMechanism` (and making a PR for this to be added to biolink) is also acceptable. But @cmungall please correct me if I'm wrong.

deepakunni3: @patrickkwang This looks good to merge. 

In the absence of a valid Biolink Model class it is okay to mint them, as long as there are requests in the Biolink Model repo to add them to the model (for ease of tracking).

patrickkwang:
Not all current Biolink slots obey the snake_case convention: https://github.com/biolink/biolink-model/pull/419

But the ones that don't are not used by biological entities, so I'm not concerned about moving forward with this while that PR is reviewed.

PR #139, opened 2020-08-05
Add alternative bypass_cache

edeutsch:
Instead of having bypass_cache as a Query object attribute, see if it is more palatable as query parameter.
Is it possible to have query parameter alongside a requestBody? I assume so, but I don't know this to work
For the `extended` branch
The assumption is that if a server does not implement this, then there is no caching. If you implement caching, you should do it this way.

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -75,11 +75,11 @@ paths:
       parameters:
         - name: bypass_cache
           in: query
-          description: Set to true in order to bypass any possible cached result message and try to answer the query over again
-          required: false
-          type: boolean
-          default: false
-          example: true
+          description: >
+            Set to true in order to bypass any possible cached result
+            message and try to answer the query over again
+          schema:
+            type: boolean
       requestBody:
         description: Query information to be submitted
         required: true
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -78,9 +78,9 @@ paths:
           description: >
             Set to true in order to bypass any possible cached result
             message and try to answer the query over again
-          default: false
           schema:
             type: boolean
+            default: false
       requestBody:
         description: Query information to be submitted
         required: true
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -72,6 +72,14 @@ paths:
       summary: Query reasoner via one of several inputs
       description: ''
       operationId: query
+      parameters:
+        - name: bypass_cache
+          in: query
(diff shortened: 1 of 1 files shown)


API/TranslatorReasonersAPI.yaml
```diff
@@ -72,6 +72,14 @@ paths:
       summary: Query reasoner via one of several inputs
       description: ''
       operationId: query
+      parameters:
+        - name: bypass_cache
+          in: query
+          description: Set to true in order to bypass any possible cached result message and try to answer the query over again
+          required: false
```
patrickkwang, line 79: `required` is false by default and could be removed here.

API/TranslatorReasonersAPI.yaml
```diff
@@ -72,6 +72,14 @@ paths:
...
+          type: boolean
+          default: false
+          example: true
```
patrickkwang, line 82: `example` feels unnecessary for boolean parameters.

API/TranslatorReasonersAPI.yaml
```diff
@@ -72,6 +72,14 @@ paths:
       summary: Query reasoner via one of several inputs
       description: ''
       operationId: query
+      parameters:
+        - name: bypass_cache
+          in: query
+          description: Set to true in order to bypass any possible cached result message and try to answer the query over again
+          required: false
+          type: boolean
```
patrickkwang, line 80: `type` needs to be enclosed in a `schema`. See [examples](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#parameter-object-examples).

edeutsch:
fixed issues described above in pull request

API/TranslatorReasonersAPI.yaml
edeutsch: fixed

API/TranslatorReasonersAPI.yaml
edeutsch: removed.

API/TranslatorReasonersAPI.yaml
edeutsch: fixed

API/TranslatorReasonersAPI.yaml
```diff
@@ -72,6 +72,14 @@ paths:
       summary: Query reasoner via one of several inputs
       description: ''
       operationId: query
+      parameters:
+        - name: bypass_cache
+          in: query
+          description: >
+            Set to true in order to bypass any possible cached result
+            message and try to answer the query over again
+          schema:
```
patrickkwang, line 81:
```suggestion
          default: false
          schema:
```
We need a default since this is a non-required parameter.

Issue #132, opened 2020-07-28
Exception Handling considered critical


webyrd:
In a production system it isn't unusual for 80% of the implementation of critical functions or methods to be exception and error handling code.

If I understand correctly, there is currently no standard way to report exceptional conditions, other than an error code mechanism or the logging mechanism.

This sounds like a giant hole, for building any production system, especially involving as many components or teams as we have.

Is there some minimal exception reporting mechanism we want to standardize on?

edeutsch:
What we do in ARAX is use the `extended` schema properties Message.message_code and Message.code_description:
```
        message_code:
          type: string
          example: OK
          description: >-
            Set to OK for success, or some other short string to indicate and
            error (e.g., KGUnavailable, TermNotFound, etc.)
        code_description:
          type: string
          example: 9 results found
          description: >-
            Extended description denoting the success or mode of failure in the
            generation of the message
```

Our system is currently returning things like this:
```
message_code	code_description
OK	9 results found
QueryGraphInterpreterUnsupportedGraph	QueryGraphInterpreter finds more nodes than supported in this QueryGraph
QueryGraphZeroNodes	QueryGraph has 0 nodes. At least 1 node is required
QueryGraphTooManyEdges	QueryGraph may not have edges if there is only one node
QueryGraphEdgeWithNoId	QueryGraph has a edge with no id. This is not permitted
NoResults	No paths were found in {kp_to_use} satisfying this query graph
InvalidQuery	Cannot expand an edge for which neither end has any curies. (Could not find curies to use from a prior expand step, and neither qnode has a curie specified.)
```
We do not have a dictionary of message_codes, but that seems possible and sensible.

patrickkwang:
@webyrd What features are we looking for from this exception reporting? HTTP status codes seem the natural mechanism for reporting errors, and I had been assuming that everyone would use them. But maybe you're looking for standardization on finer details than `400 Bad Request` or `500 Internal Server Error`?

webyrd:
@patrickkwang I'm thinking of something much more fine grained, that would allow for *recovery* from exceptional states.  @edeutsch has some nice examples in his response above, such as TermNotFound or KGUnavailable.

> We do not have a dictionary of message_codes, but that seems possible and sensible.

That sounds like a reasonable start.

patrickkwang:
[...]

PR #140, opened 2020-08-05
Add explicit HTTP status codes

edeutsch:
Provide a complete set of HTTP status codes and their meanings and contents

### DIFF
--- a/API/TranslatorReasonersAPI.yaml
+++ b/API/TranslatorReasonersAPI.yaml
@@ -69,44 +69,36 @@ paths:
               $ref: '#/components/schemas/Query'
       responses:
         '200':
-          description: Successful operation
+          description: >-
+            OK. There may or may not be results. Note that some of the provided
+            identifiers may not have been recognized.
           content:
             application/json:
               schema:
                 $ref: '#/components/schemas/Message'
         '400':
           description: >-
-            Bad request. One or more specified
-            parameters or objects are not valid
+            Bad request. The request is invalid according to this OpenAPI
+            schema OR a specific identifier is believed to be invalid somehow
+            (not just unrecognized).
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Message'
-        '404':
-          description: >-
-            No results found for the specified
-            parameters and objects
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/Message'
+                type: string
         '500':
           description: >-
-            Server error. The specified parameters and
-            objects may be valid but the server was
-            unable to service the request
+            Internal server error.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Message'
+                type: string
         '501':
           description: >-
-            Not implemented. This end point is not
-            implemented by this resource
+            Not implemented.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Message'
+                type: string
       x-swagger-router-controller: swagger_server.controllers.query_controller
 components:
   schemas:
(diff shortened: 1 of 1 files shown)


CaseyTa:
If we receive a valid query for which we find no results, should we return an empty KG with status 200 or 404? 

Should there be a distinction in responses between 
1. A Translator component has the resources to operate on a valid query but finds no results
1. A Translator component does not have the resources (e.g., the requested nodes or edge types are not in the component's KG) to operate on the given query?

patrickkwang:
Note the [tests](https://travis-ci.com/github/NCATS-Tangerine/NCATS-ReasonerStdAPI/builds/178486545) are failing because several of the `description`s are too long.
Use multiline strings: https://stackoverflow.com/a/21699210/13242055

patrickkwang:
I think it's a good idea to define all of these codes, but I don't think the Message is the right payload for most of them. 404 500 and can probably just have a `type: string` body, and we should discuss the correct form for the 400 body, to address #132.

edeutsch:
fixed the description lines in the revised PR

edeutsch:
> If we receive a valid query for which we find no results, should we return an empty KG with status 200 or 404?

Although I am not fond of it, I think general best web services practice is to return a 404.
(thus making it difficult to distinguish between a valid query with no good answer and hitting an endpoint with /qeury misspelled)

edeutsch:
> Should there be a distinction in responses between
>
>   1. A Translator component has the resources to operate on a valid query but finds no results
>   2. A Translator component does not have the resources (e.g., the requested nodes or edge types are not in the component's KG) to operate on the given query?

I do not feel it is necessary to have different HTTP codes for such nuances. These would both be 404. The way we report such differences (using the existing `extended` schema constructs) is:
- Message.message_code contains different strings such as 'NoSuitableAssociationsFound' or 'NoSuitableKPsFound'
- Message.code_description contains a human-readable sentence describing the problem
- Message.log contains a trail of logging showing how things were going right until things started to go bad so that a user/tester can understand the problem a bit better.

We could have a common agreed upon set of Message.message_codes documented somewhere.

patrickkwang:
I think "no results" should be a 200. Imagine a sort of "is it true that..." question in which you have two nodes with specified CURIEs connected by an edge with specified predicate. The answer to the implied question may be "no". In that case, no server error has occurred, and the server correctly responds that it can find no evidence for this assertion. 0 results is just an extreme case of N results and should not be treated differently by the server. As @edeutsch points out, 404 is for cases like misspelled URLs where the server actually fails to fulfill the client's request.

Providing feedback about _why_ no results were found is tricky, especially if the server is an ARA. Is there a good reason to treat no-results cases differently than few-results cases? Either way, the client is likely to ask "why didn't I get more?" Rather than worry about delivering concise "error" codes, I'd think about how servers can provide a more detailed description of the steps they took, including things like "KP1 does not accept this predicte", "KP2 did not recognize this CURIE", and "KP3 happily returned 0 results."

webyrd:
I agree with @patrickkwang's last comment.  No results is different than an erroneous query, or a query in which an exceptional condition occurred.

> Rather than worry about delivering concise "error" codes, I'd think about how servers can provide a more detailed description of the steps they took, including things like "KP1 does not accept this predicte", "KP2 did not recognize this CURIE", and "KP3 happily returned 0 results."

I agree.

rtroper:
A 404 on a valid query with no results could create a red herring, leading the user on a fruitless effort to figure out what's wrong with their query. I definitely advocate 200 in this case.

PR #141, opened 2020-08-05
Edit the top-level README with current information

edeutsch:
Refine the top-level README to provide current information

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## Project Summary
+# Translator Reasoner API
 
 The current NCATS Translator Reasoner API is defined by an OpenAPI YAML file in this repository.
 The core schema is defined in the `master` branch. An extended schema is defined in the `extended` branch.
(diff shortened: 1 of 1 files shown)


CaseyTa:
The 2 Google Docs (development agenda and high level summary) are not publicly accessible. 

I don't think we should make the Agenda notes publicly available unless we at least remove the Zoom link with the embedded password. There may be other information in the agenda that we don't want shared publicly. 

I think the summary document should be fine to make public.

patrickkwang:
I think we should avoid having any links to Google Docs. Not only do you run into permissions issues, but documents there just feel too ephemeral. Anything critical we should copy into Markdown documentation in this repo.

edeutsch:
I have updated the pull request to remove links to Google docs and made other cosmetic changes. Please re-review

patrickkwang:
Do we not want a top-level header that says "Translator Reasoner API"? It's basically the title of the homepage of the project.

edeutsch:
okay, you're right. Fixed. That looks better.

PR #147, opened 2020-09-03
Add asynchronous interface

patrickkwang:
`/aquery` returns a query_id (string)
`/aresponse/{query_id}` returns a Message

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -100,6 +100,97 @@ paths:
               schema:
                 type: string
       x-swagger-router-controller: swagger_server.controllers.query_controller
+  /aquery:
+    post:
+      tags:
+        - query
+      summary: Asynchronously query reasoner via one of several inputs
+      description: ''
+      operationId: aquery
+      requestBody:
+        description: Query information to be submitted
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/Query'
+      responses:
+        '202':
+          description: >-
+            Accepted. Poll /aresponse for results.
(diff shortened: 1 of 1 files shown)


patrickkwang:
Discussion points:
1. naming of endpoints
2. status reporting - What should happen if query is still processing?

TranslatorReasonerAPI.yaml
```diff
@@ -100,6 +100,97 @@ paths:
...
+          description: Query id returned from /aquery
+      responses:
+        '200':
+          description: >-
+            OK. There may or may not be results. Note that some of the provided
+            identifiers may not have been recognized.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Query'
```
edeutsch, line 171: Should be
   $ref: '#/components/schemas/Message'
instead?

edeutsch:
> 2. status reporting - What should happen if query is still processing?

My suggestion: return a minimal message with a message_code of "Processing":
{ "message_code": "Processing", "code_description": "Submitted query is still being processed" }

> (3. error reporting - What should happen if the query failed?)

My suggestion: return a minimal message with error information:
{ "message_code": "QueryGraphInterpreterUnsupportedGraph", "code_description": "QueryGraphInterpreter cannot interpret this QueryGraph", "log": [ "Logging information that might tell more about what went wrong" ] }

vdancik:
Alternative endpoints: 

- `/query/submit`
- `/query/status`
- `/query/response`

vdancik:
milestone: 1.1

edeutsch:
Closed in favor of #255

Issue #92, opened 2020-01-29
tighten up `required` and `additionalProperties`


patrickkwang:
Making desired properties optional and additional properties permitted leads to perhaps puzzling behavior. Mis-named properties are then not validated at all.

For example, [this](https://github.com/NCATSTranslator/KnowledgeProviderScaffold/blob/4b5ee43956fce22e295e3bd62920419e19fc06d8/TranslatorKnowledgeProviderResponseScaffold.json) validates against the 0.9.2 spec, despite providing `answers` instead of `results`, and formatting the contents differently.

edeutsch:
I agree.
One option would be to remove additionalProperties from everywhere, have a good ongoing discussion about what we will permit in the (maximal) _optional form of the standard, and then validate against the optional form. That way, if someone wants to add something, it will be a group discussion about if adding it is reasonable and that there's only one way to do something (i.e. not that two different groups do the same thing in two different ways). And anything not in the maximal version is illegal. All servers would still be only obligated to implement the minimal version and then pick and choose from optional.
Does that sound like a good strategy?

patrickkwang:
That's a reasonable approach to validating the various JSON components.

Note that we will have to explicitly specify `additionalProperties: false` in the optional document, since `additionalProperties: true` is actually the default behavior.

patrickkwang:
#153

PR #148, opened 2020-09-03
Response metadata

patrickkwang:
This slightly reframes how we talk about requests and responses by replacing the Query and Message structures with Request, Response, and Data. It also adds structured `logs` as metadata to the Response.

Formerly, requests to /query were Query objects, each of which contained a Message. We basically rename these, referring to them as Request and Data respectively, where Data is identical to Message except that extra fields beyond `query_graph`, `knowledge_graph`, and `results` are not allowed.

Formerly, responses from /query were Message objects. Now they are Response objects, each of which contains Data. Then a list of LogEntries is added to serve as response "metadata". Each LogEntry includes fields `timestamp`, `level`, `message` (not to be confused with the old usage), and `code` (which serves as a standardized label e.g. QueryNotTraversable, KPNotAvailable, or KPResponseMalformed).

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -66,7 +66,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/Query'
+              $ref: '#/components/schemas/Request'
       responses:
         '200':
           description: >-
@@ -75,7 +75,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Message'
+                $ref: '#/components/schemas/Response'
         '400':
           description: >-
             Bad request. The request is invalid according to this OpenAPI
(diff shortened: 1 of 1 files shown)


patrickkwang:
Addresses #132.

patrickkwang:
Fixes the specific issue described in #92, wherein the (allowed) field `answers` replaces the (not required) field `results`.

ehinderer (approved):

edeutsch (commented):
The logs and logEntry is fine. This is more complex than what we have in `extended` but is highly similar to what we already do, so supporting this is relatively easy.

I am hesitant about the Query->Request and Message->Response change. It's a fine way to do it. But is it better? Do we want to put ourselves through this pain of rewriting our classes and all the code that uses them for this? What do we gain by renaming Message to Data? Data is even more generic and bland than message. Do we want this pain?

patrickkwang:
Closing in favor of #153.

PR #150, opened 2020-09-04
Adding a field for a callback URL

MarkDWilliams:
Not sure if I made these proposed changes correctly.  If not, please feel free to amend.  The general idea is to provide a callback option for longer running queries to avoid need to leave the connection open while computation takes place.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -109,7 +109,9 @@ components:
         message:
           $ref: '#/components/schemas/Message'
         callback:
-          type: url
+          type: string
+          format: uri
+          pattern: ^https?://
       additionalProperties: true
       required:
         - message
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -108,8 +108,8 @@ components:
       properties:
         message:
           $ref: '#/components/schemas/Message'
-	callback:
-		type: url
+        callback:
+          type: url
       additionalProperties: true
       required:
         - message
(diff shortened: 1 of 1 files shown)


TranslatorReasonerAPI.yaml
```diff
@@ -108,6 +108,10 @@ components:
       properties:
         message:
           $ref: '#/components/schemas/Message'
+        callback:
+          type: string
+          format: uri
+          pattern: ^https?://
```
patrickkwang, line 114:
```suggestion
          pattern: ^https?://
          description: >-
            Upon completion, a POST request will be made to this URL.
            The request body will be a JSON-encoded Message.
```

vdancik:
This pull request should be merged with #147 into a single pull request.

edeutsch:
Presumably there needs to be some way for the client to identify the query unless they are always done serially (i.e. the client never submits another query until the previous one completes). Perhaps this can just be accomplished in the URL. For example, the callback url string might be:
http://arax.rtx.ai/api/rtx/v1/message_collector/28384
where /message_collector is the endpoint and the 28384 is the client's id of the request
Is that how this is envisioned?

Also, in the current design, this means that callback could be used in /query or /aquery?

edeutsch:
Didn't we decide that parameters such as this (that have nothing to do with the content of the query) should go into query (i.e. URL) parameters rather than in the payload?

vdancik:
milestone: 1.1

edeutsch:
Closed in favor of #255

PR #151, opened 2020-09-08
Move edge relation from extended to core API

vdancik:
According to latest data-modeling discussion it is expected to provide both, an original low-level relation and a Biolink-compliant predicate.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -241,6 +241,10 @@ components:
             - type: array
               items:
                 $ref: '#/components/schemas/BiolinkRelation'
+        relation:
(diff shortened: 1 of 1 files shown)


ehinderer (approved):

vdancik:
milestone: 1.0

edeutsch (approved):
Fine with me. This appears to be exactly as implemented in extended. We already support it to some degree based on extended. Fine to move to core. One could argue that it alters the intent of the query and therefore should not languish in extended.

patrickkwang (approved):

CaseyTa (approved):

bill-baumgartner (approved):

Issue #105, opened 2020-06-11
Allow logical operators in query graph


patrickkwang:
A query graph containing edges `e0`, `e1`, and `e2` is traditionally interpreted as implying that all assertions are true, i.e. `e0 AND e1 AND e2`. As a side note, it would be good to make that interpretation explicit somewhere. But that raises the question of how to represent different and more complex logical operations, e.g. `e0 OR (e1 AND e2)` or `(e0 OR e1) AND NOT e2`.

This is tied up with data modeling issues around polarity and strength of evidence for individual assertions (relationships/edges). We should be clear about what it means to bind some assertion onto `e0`. I think it ought to mean just that there is evidence for the assertion (and assertion matches the query template). Note that the assertion may be of the form e.g. "A does not cause B", which would be modeled with a `causes` predicate and `negated = True`. Then binding that assertion indicates "we know that A does not cause B". Binding `NOT e0` would indicate a lack of evidence for the assertion, i.e. "we do not know that A does not cause B". It would then be up to the client whether they wish to assume that absence of evidence is evidence of absence ("A causes B"), but the server would make no such assumption.

In addition to agreeing on an interpretation of these logical operations, we need to determine how to represent them, syntactically. That should be easier.

patrickkwang:
I'd welcome your feedback, @cmungall @mbrush

webyrd:
Does this tie into handling negated edges, in the extended standard?

patrickkwang:
> Does this tie into handling negated edges, in the extended standard?

Yes. Anything that changes the interpretation of an edge will need to end up in the core standard.

ehinderer:
Would the responsibility fall on the ARAs to be able to interpret this logic?

patrickkwang:
> Would the responsibility fall on the ARAs to be able to interpret this logic?

Yes.

cmungall:
There is a proposed data model for this in https://github.com/biolink/biolink-model/pull/414

Although for convenience I made this as part of the model, this doesn't preclude a parallel json-schema on the reasonerapi side of things. There are some interesting semantic implications for rdf graphs in making disjunctions of assertions part of the domain of discourse.

nlharris:
@patrickkwang do you have what you need to resolve this issue now?

patrickkwang:
I'd like to see https://github.com/biolink/biolink-model/pull/414 finalized, then we need to draft a pull request here which, when merged, will close this issue.

PR #157, opened 2020-09-29
Expand query node/edge property constraints

patrickkwang:


stevencox (approved):

vdancik (changes requested):
instead of 'additional properties' I would use 'where' to indicate that these are query constrains.

edeutsch:
I am having some difficulty envisioning that the JSON is going to look like from staring at the YAML. Do we have some JSON snippets that display the proposed examples?

patrickkwang:
Here's an example snippet:
```yaml
query_graph:
  nodes:
    n0:
      categories:
      - biolink:Gene
      location:
        operator: matches
        value: 11q13.*
      gene_family:
        operator: in
        value:
        - BBSome
        - Antisense RNAs
    n1:
      categories:
      - biolink:Disease
      congenital_abnormality: true
    n2:
      categories:
      - biolink:ChemicalSubstance
      mass:
        operation: '>'
        value: 57.0
```

CaseyTa:
On the topic of strict vs loose interpretation of constraints, given that we're already having trouble within the Fix-It sessions to get results and that many people are starting off their queries with general queries and then modifying their queries based on what they're finding, would it be easier to allow results to be returned if they loosely meet the constraints? For example, a KP can return a result if the result is not known to violate any of the supplied constraints. If a loose interpretation of constraints is allowed, perhaps we could also provide a mechanism within the result bindings to indicate which constraints were understood and satisfied per result. In order for a strict interpretation to work, it seems that an ARA would have to be very familiar with the information a KP has and which constraints it can operate on. 

Numeric operations would need a property for units. 

Does the discussion of how the constraints are identified (e.g., gene_family, location in Patrick's example) and their values fall under this pull request or is that a separate topic?

yakaboskic (changes requested):
Following @CaseyTa, I think before something like this is implemented KP/ARAs need to have requirements/options laid out somewhere of what constraints they support. I am not sure if this should be done here or somewhere else. It's a cool idea tho and I think I like it better than additional properties.

andrewsu (approved):

bill-baumgartner (approved):
I approve these changes as long as the Service Provider (@kevinxin90) is on board as well since they are hosting our KP.

marcdubybroad (approved):

brettasmi (approved):
These changes look good to me. I think the syntax changes are pretty clear. I might prefer `==` to `=`, but that's not a big deal.

This does seems like a lot to functionally implement by 1.1 without some other specifications elsewhere in Translator, but I don't think we need to worry about to get the syntax in place.

brettasmi:
Personally, I'd prefer a stricter interpretation of the specified constraints, or at least standard guidance on how we can "loosely" interpret them. I think it's the responsibility of the client to specify a query that the service instead of allowing the service to do some non-standardized alteration of the contract.

rtroper (approved):

edeutsch (commented):
I have several concerns about this:

1. this is a huge OneOf which many struggled with last time

2. Where do we define centrally what each property name means?

3. Can we consider replacing the property name with a concept CURIE?
E.g. mass → EDAM:data_0844 (molecular mass)
E.g. location → NCIT:C13432 (Chromosome band)
This probably does not absolve us of having a central table of recognized context constraints, though

And then also implementation aspects that concern me:
4. Do ARAs just blindly pass these along to their KPs?

5. Do KPs error out if they do not recognize a property? If they ignore a property, will they be answering a different question than was asked?

6. This seems mighty complex to me. Can we come up with a concrete example that a KP is ready to implement next month and chew on that for a bit?

7. If this goes in the Core, what are everyone's obligations (see items 4,5)

patrickkwang:
Re: loosely vs. strictly interpreting constraints, I agree this needs to be spelled out. I think that decision/specification falls to Operations and Workflow. Likely we can add a flag to the appropriate operation(s) that specifies whether the server is allowed ignore constraints it cannot evaluate.

patrickkwang:
There's an additional complication. Representing simple constraints like 'mass > 57' is not too bad:
```json
{"operator": ">", "value": 57.0}
```

If we wanted to represent 'mass > 57' OR 'mass < 45', we could stick them in an array:
```json
[
    {"operator": ">", "value": 57.0},
    {"operator": "<", "value": 45.0}
]
```

But what if we want to represent 'mass > 57' AND 'mass < 65'? You could go off the deep end here and write really complex constraints. Attribute constraints could be conditional on the values of _other attributes_... At that point we're encroaching on other territory: #105 

So my inclination is to continue to treat arrays as disjunctions, like we do with other constraints (categories, ids, predicates), and put off the more complex logic until we have pressing use cases for it. In the meantime, we can actually repurpose `in` to mean "between", since the array will cover "in set".

edeutsch:
Seems reasonable. I don't quite understand how we can 'repurpose in to mean "between"' though.
It seems like we could have an explicit "between" operator that takes a list of two.
While "in" should be a set of [1..n] equalities?

I was also thinking that "not" could be a modifier to all constraints rather than being its own option (hmm, I just looked at the latest PR and "not" appears to be missing now...?) If "not" is a modifier for any of these, then you could do "not in", "not between", "not ==", "not matches", etc. Seems like any of the constraints could be reasonably negated. (and that mechanism could in principle remove the need for <>, >=, <=)

patrickkwang:
A set of options is handled like this:
```json
gene_family:
- BBSome
- Antisense RNAs
```

which is nice because it reflects how we're already handling categories, predicates, and ids.

If we take `in` to mean "in range",
```json
mass:
  operator: in
  value: [45, 57]
```

means "mass between 45 and 57".

I don't think this PR ever provided a "not", but it's a good idea. I'd suggest we put it off, however, until we figure out whether/how to handle all sorts of complex logic e.g. "(A and not B) or (C xor D)"

edeutsch (commented):
On a technical level, can you check the indentation of additionalProperties? It is indented one level deeper than before and I think it was correct before? It needs to be at the same level as properties? I'm not sure.

stevencox (approved):

stevencox:
I do like @vdancik's suggestion about using "where" - it's familiar terminology across query languages.

There are a lot of good comments above indicating a desire for more definition on how this should work in various ways. My thoughts:
* On strict vs loose, I hear the concern about making Fix-it sessions work. Still, I like strict interpretation of operators. It'll make the resulting system more predictable and understandable. We can specify permissive ranges or not specify constraints if result sets are too limited.
* On passing constraints to KPs, suppose 
  * ARAs pass all constraints they receive along to KPs. 
  * If a KP can't handle a constraint, it throws an error
  * ARAs decide if that's catastrophic or not.
  * ARAs provide a clear error message detailing which KP errored and on which constraint.
  * What does this break?
* @edeutsch you mentioned curies and some syntactic issues. @balhoff mentioned representing constraints with predicates. Do folks want to propose specific alternative representations? Personally, I'm ok with @patrickkwang's syntax above. 
* As to specifying possible values - this can be done in the OpenAPI with enums and appropriate documentation, right?

patrickkwang:
My beef with putting this all in a "where" or "constraints" field is that we _already have_ an implied WHERE clause, which is exactly what our `categories`, `predicates`, and `ids` fields do. The proposed additional properties behave exactly the same way as the standardized fields we already have, except they can be more expressive with the operator/value syntax.

stevencox:
Yep, fair enough.

edeutsch:
From my perspective the #222 alternative has the same functionality but avoids a thicket of oneOfs and avoids arbitrarily named properties, both of which I fear will cause big headaches for groups trying to make a class-based data model out of TRAPI. And it is consistent with what we have done with Attribute. I view consistency with other parts of the model and avoiding potential pitfalls that afflicted us in the past more compelling than a sense that QNodes and QEdges are already implicitly a WHERE.

edeutsch:
Sorry, I'm catching up on many threads sometimes a bit out of order. Regarding:

> Do folks want to propose specific alternative representations?

PR #222 is the proposed alternative. Please review it.

We have 1 week to decide on either #157 or #222

We will discuss at tomorrow's call.

edeutsch:
One more response to @stevencox comment:

> •	On passing constraints to KPs, suppose 
>o	ARAs pass all constraints they receive along to KPs.
>o	If a KP can't handle a constraint, it throws an error
>o	ARAs decide if that's catastrophic or not.
>o	ARAs provide a clear error message detailing which KP errored and on which constraint.
>o	What does this break?

I think this is right and fine. The key component is:

>o	If a KP can't handle a constraint, it throws an error

I think this is right. But I think during earlier discussion of this there was some sentiment that alternatively KPs can just ignore constraints they don't understand. I am concerned that would lead to KPs answering different questions than are asked.

So I think this (ignore or error) is an implementation expectation that should be documented clearly (can be after Apr 1)

edeutsch:
Although with considerable misgivings by some team members, it was decided 2021-03-25 that for TRAPI 1.1 we would proceed with PR #222. The concept of providing names for constraints as first-class properties for both Attributes and Constraints (together) may be revisited at some later time. Closing for now.

patrickkwang:
This is very similar to how constraints are implemented in MongoDB queries: https://docs.mongodb.com/manual/tutorial/query-documents/

edeutsch:
Result of discussion on call on 2021-07-22: Let it age for now and remain open, to be re-examined during TRAPI 1.3 phase.

kshefchek (approved):

PR #163, opened 2020-10-15
realign with biolink model terms

edeutsch:


TranslatorReasonerAPI.yaml
```diff
@@ -322,7 +322,7 @@ components:
           type: string
           example: Haptoglobin
           description: Formal name of the entity
-        type:
+        category:
           oneOf:
             - type: string
```
patrickkwang, line 327:
```suggestion
            - $ref: '#/components/schemas/BiolinkEntity'
```
This is just an unrelated fix I noticed during review.

vdancik (approved):

patrickkwang (approved):

CaseyTa (approved):

rtroper (approved):

TomConlin:
+1

bill-baumgartner (changes requested):
The current changes look good. There are potentially two more changes required to align with biolink (apologies if there is a better way to point these out.. I couldn't figure out how to highlight them without creating new issues):

1. `QNode` has a property called `curie` (line 262). To align with the KGX format I think this should be renamed to `id`.
2. `Node` has two properties defined: `name` & `category` (lines 320-330). In KGX, a node has two required properties: `id` & `category`. The `name` property is optional ([see KGX docs](https://github.com/biolink/kgx/blob/master/data-preparation.md)) so I think we need to add an `id` property under `Node`. Is there a reason `Node` does not have an explicitly defined `id` property?

vdancik:
@bill-baumgartner,
 
> 2. `Node` has two properties defined: `name` & `category` (lines 320-330). In KGX, a node has two required properties: `id` & `category`. The `name` property is optional ([see KGX docs](https://github.com/biolink/kgx/blob/master/data-preparation.md)) so I think we need to add an `id` property under `Node`. Is there a reason `Node` does not have an explicitly defined `id` property?

Actually nodes (and edges) do have id's, they are specified as keys of knowledge_graph.nodes and knowledge_graph.edges.

patrickkwang:
> 1. `QNode` has a property called `curie` (line 262). To align with the KGX format I think this should be renamed to `id`.

This is interesting. As @vdancik pointed out, we've converted all query_graph and knowledge_graph nodes and edges from lists to maps and made the ids keys. Previously, we had a required `id` (e.g. "n0") on qnodes in addition to the optional `curie` (e.g. "MONDO:0100096"). Now that the don't have the `id` property, it may make sense to rename `curie` to `id` to reflect the knode property that it constrains. This would align nicely with #157. But it could cause confusion when talking about these things because qnodes could have two "ids" - the unique identifier for the qnode itself, and the id of the knode to which it must be bound. Though it was a confusing before, too. Ultimately, I think I'm in favor of changing `curie` to `id`.

bill-baumgartner:
> Actually nodes (and edges) do have id's, they are specified as keys of knowledge_graph.nodes and knowledge_graph.edges.

@vdancik Thank you for the clarification. I remember this discussion now. The descriptions for the `nodes` and `edges` properties for both `KnowledgeGraph` and `QueryGraph` still refer to "Lists" however. Should these instead be described as maps? Perhaps this is a task for the upcoming documentation effort.

brettasmi (approved):

PR #164, opened 2020-10-19
Rename qnode.curie -> id

patrickkwang:
https://github.com/NCATSTranslator/ReasonerAPI/pull/163#pullrequestreview-509705011

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -201,19 +201,19 @@ components:
     NodeBinding:
       type: object
       properties:
-        kg_id:
-          type: string
+        id:
+          $ref: '#/components/schemas/CURIE'
           description: A knowledge-graph node id, i.e. the `id` of a KNode
       required:
-        - kg_id
+        - id
     EdgeBinding:
       type: object
       properties:
-        kg_id:
+        id:
           type: string
           description: A knowledge-graph edge id, i.e. the `id` of a KEdge
       required:
-        - kg_id
+        - id
     KnowledgeGraph:
       type: object
       description: >-
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -261,10 +261,10 @@ components:
       properties:
         id:
           oneOf:
-            - type: string
+            - $ref: '#/components/schemas/CURIE'
             - type: array
               items:
-                type: string
+                $ref: '#/components/schemas/CURIE'
           example: OMIM:603903
           description: CURIE identifier for this node
         category:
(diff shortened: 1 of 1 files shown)


bill-baumgartner (approved):

vdancik (approved):

edeutsch:
My concern with this is that it will cause confusion. QNode used to have both:
- id  (e.g. n00)
- curie (e.g. UniProtKB:P12345)
We then moved the id into the map key, which ensures uniqueness and faster lookups. But it's still the id, isn't it?
If we rename the curie to id, what do we say to refer to what used to be the id (e.g. n00)?

Do we say that the key is n00 and the id is UniProtKB:P12345 ?

vdancik:
+1 to 'key' and 'id' terminology.

vdancik (changes requested):
Actually there is the same issue in the EdgeBinding/NodeBinding classes where we have `kg_id` field. Since we migrated query keys from the Binding classes, those classes are not necessary anymore and can be replaced with strings.

edeutsch:
> Actually there is the same issue in the EdgeBinding/NodeBinding classes where we have `kg_id` field. Since we migrated query keys from the Binding classes, those classes are not necessary anymore and can be replaced with strings.

Except that part of the intent of breaking them out separately like this was to enable the possibility of adding an additional property on these bindings. We don't do this yet, but I think Robokop was providing some score on the *binding*, that might be different from result to result for the same KG edge.

patrickkwang:
Right, @edeutsch, we wanted the option to attach metadata (like scores) to bindings.

Probably `kg_id` could be renamed, too, since "id" will be reserved exclusively for the knowledge graph ("key" for the query graph).

edeutsch:
I am concerned that we are going out of our way to make the attribute names as cryptic as possible, changing attribute names that were fairly clear what they are into ones that are inscrutable without reading the docs.

webyrd:
> I am concerned that we are going out of our way to make the attribute names as cryptic as possible, changing attribute names that were fairly clear what they are into ones that are inscrutable without reading the docs.

I agree.  The TRAPI standard is built with BioLink Model in mind, correct?  In that case, clearly distinguishing between the CURIE and the node id seems like a good thing for readability and avoiding confusion.

vdancik (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -381,12 +381,12 @@ components:
           example: upregulates
```
bill-baumgartner, line 381:
```suggestion
          $ref: '#/components/schemas/CURIE'
          example: RO:0002213
```
The `relation` should also be constrained to a CURIE, correct? Note that the suggested changes above should also remove `type: string` on line 380 however GitHub won't allow me to comment on that line directly.

patrickkwang: I don't know if `relation` is meant to be a CURIE; we should just follow the lead of the data modeling committee here. Is there a place that those sorts of decisions are recorded? Regardless, that can be a separate PR.

edeutsch (approved):
Adding a CURIE type definitely seems like a plus, that makes things clearer. Changing curie to id and kg_id to id makes things more cryptic. I will understand it, but it will make it more confusing for new users.

PR #167, opened 2020-10-22
Update documentation for Response object

CaseyTa:
Updated descriptions and examples for the Response object.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -113,6 +113,13 @@ components:
         - message
     Response:
       type: object
+      description: >-
+        The Response object contains the main payload when a TRAPI query 
+        endpoint interprets and responds to the submitted query successfully 
+        (i.e., HTTP Status Code 200). The message property contains the 
+        knowledge of the response (query graph, knowledge graph, and results).
+        The status, description, and logs properties provide additional details 
+        about the response. 
       properties:
         message:
           $ref: '#/components/schemas/Message'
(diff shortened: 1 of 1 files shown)


CaseyTa:
I have a few lingering questions:

1. Are the allowed status codes documented somewhere? If not, should we enumerate them here? Are they the same set of codes allowed for LogEntry's codes? 
1. Should I create an example for the Response object? I opted not to do so for now since it would be long.

edeutsch:
I think the status codes should be documented elsewhere. but I would pick one of the error conditions at:
https://docs.google.com/document/d/12GRjcAqXQfp557kAcVEm7V0mE3hKGxAY5B6ZPavp6tQ/edit
as the example.

The message property probably should have a description, too.

I don't think the Response object itself needs an example.

CaseyTa:
> I think the status codes should be documented elsewhere. but I would pick one of the error conditions at:
> https://docs.google.com/document/d/12GRjcAqXQfp557kAcVEm7V0mE3hKGxAY5B6ZPavp6tQ/edit
> as the example.

I considered that, but I figured that since we wanted examples within a class to be compatible, that we might also want examples to be cohesive throughout the yaml (as much as possible). I figured a 'success' example would be easier to be cohesive on. I'm happy to change this to an error condition if that's still preferred.  
 
> The message property probably should have a description, too.

I initially didn't think that was allowed since [sibling elements of $refs are ignored](https://swagger.io/docs/specification/using-ref/), but we're already adding descriptions next to $refs for other properties (e.g., the query_graph and knowledge_graph properties in Message), so I went ahead and added the description for the message property. 

> I don't think the Response object itself needs an example.

It might be nice to have a folder with examples of complete messages for queries and responses. This would also allow us to have multiple examples for different types of operations and possibly different versions of the spec.

patrickkwang (approved):

bill-baumgartner (approved):

vdancik (approved):

edeutsch (approved):

PR #170, opened 2020-10-27
QEdge, Edge, edge_bindings and EdgeBinding documentation

RichardBruskiewich:
Also commented out suspected bug in `edge_bindings` value specification?

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -263,7 +263,6 @@ components:
       additionalProperties: true
       required:
         - id
-      additionalProperties: true
     KnowledgeGraph:
       type: object
       description: >-
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -192,9 +192,12 @@ components:
           type: object
           description: >-
             Dictionary of Input Query Graph to Result Knowledge Graph edge
-            bindings where the dictionary keys are the key identifiers of the
+            bindings where the dictionary keys are the key identifiers of the 
             Query Graph edges and the associated values of those keys are
-            instances of EdgeBinding schema type (see below).
+            instances of EdgeBinding schema type (see below). This value is
+            an array of EdgeBinding's since a given query node may resolve to
+            multiple knowledge graph nodes with the same predicate/category
+            semantics.
           additionalProperties:
             type: array
             items:
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -198,7 +198,7 @@ components:
           additionalProperties:
             type: array
             items:
-            $ref: '#/components/schemas/EdgeBinding'
+              $ref: '#/components/schemas/EdgeBinding'
       required:
         - node_bindings
         - edge_bindings
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -190,11 +190,16 @@ components:
               $ref: '#/components/schemas/NodeBinding'
         edge_bindings:
           type: object
-          description: List of QEdge-KEdge bindings.
+          description: >-
+            Dictionary of Input Query Graph to Result Knowledge Graph edge bindings where the
+            dictionary keys are the key identifiers of the Query Graph edges and the associated values of
+            those keys are instances of EdgeBinding schema type (see below).
           additionalProperties:
-            type: array
-            items:
-              $ref: '#/components/schemas/EdgeBinding'
+            # TODO: review (and possibly remove) this
+            # (possibly superfluous) array type specification here?
+            # type: array
+            # items:
+            $ref: '#/components/schemas/EdgeBinding'
       required:
         - node_bindings
         - edge_bindings
@@ -208,10 +213,14 @@ components:
         - id
     EdgeBinding:
       type: object
+      description: >-
+        A instance of EdgeBinding is a single KnowledgeGraph Edge mapping, identified by the corresponding 'id' object
+        key identifier of the Edge within the Knowledge Graph. Future iterations of the EdgeBinding record may include
+        other annotation, such as a score to qualify the computed confidence of the mapping.
       properties:
         id:
           type: string
-          description: A knowledge-graph edge id, i.e. the `id` of a KEdge
+          description: The key identifier of a specific instance of KnowledgeGraph Edge.
       required:
         - id
     KnowledgeGraph:
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -192,7 +192,7 @@ components:
           type: object
           description: >-
             Dictionary of Input Query Graph to Result Knowledge Graph edge
-            bindings where the dictionary keys are the key identifiers of the 
+            bindings where the dictionary keys are the key identifiers of the
             Query Graph edges and the associated values of those keys are
             instances of EdgeBinding schema type (see below). This value is
             an array of EdgeBinding's since a given query node may resolve to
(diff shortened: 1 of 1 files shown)


edeutsch (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -208,10 +215,16 @@ components:
         - id
     EdgeBinding:
       type: object
+      description: >-
+        A instance of EdgeBinding is a single KnowledgeGraph Edge mapping,
+        identified by the corresponding 'id' object key identifier of the
+        Edge within the Knowledge Graph. Future iterations of the EdgeBinding
+        record may include other annotation, such as a score to qualify the
```
patrickkwang, line 222: Other annotations like these are currently allowed since `additionalProperties` defaults to `true`. They are just not currently standardized.

patrickkwang: Can we clarify this?

patrickkwang (changes requested):

TranslatorReasonerAPI.yaml
```diff
@@ -190,7 +190,14 @@ components:
               $ref: '#/components/schemas/NodeBinding'
         edge_bindings:
           type: object
-          description: List of QEdge-KEdge bindings.
+          description: >-
+            Dictionary of Input Query Graph to Result Knowledge Graph edge
+            bindings where the dictionary keys are the key identifiers of the
+            Query Graph edges and the associated values of those keys are
+            instances of EdgeBinding schema type (see below). This value is
+            an array of EdgeBinding's since a given query node may resolve to
```
CaseyTa, line 198: In the last sentence, may want to replace "query node" with "query edge", and likewise with "knowledge graph nodes"

TranslatorReasonerAPI.yaml
CaseyTa: Should these annotations all be captured in the node and edge attributes in the knowledge graph? Or are there use cases where they make more sense in the bindings?

TranslatorReasonerAPI.yaml
RichardBruskiewich: Thank you @patrickkwang and @CaseyTa, I've push a clarification of the comments for this PR based on your feedback.

CaseyTa: @RichardBruskiewich There's still two mentions of nodes in the last sentence. 

Also, I don't think the knowledge graph edges that are being bound to the same query graph edge all have to have the same predicate. If the query graph edge doesn't specify the type of the edge, then I believe any KG edge can satisfy that QG edge. Similarly, the KG edge could be different descendants of a QG edge with a specified predicate that's higher in the hierarchy.

TranslatorReasonerAPI.yaml
patrickkwang: One item that we'd like to add as an annotation on a binding is a score - How well does this kedge align with this qedge?

RichardBruskiewich: @patrickkwang, I did mention the `score` as the example of extra binding data. I have been taking this PR as mainly a secondary documentation PR but I guess as long as `score` is not currently specified as `required`, I guess there is no harm in adding it (documented) into this PR, right?

TranslatorReasonerAPI.yaml
RichardBruskiewich: Understood @CaseyTa . Making the modifications which will be committed shortly.

I have also decided to add the QEdge and Edge documentation task to this PR. I'll work on this after committing the repaired EdgeBinding text.

RichardBruskiewich:
@CaseyTa @patrickkwang @edeutsch I've just posted my "final" documentation edit for the day. I hope that it is ready to merge.

Please also do see the TRAPI semantic 'bug' PR which I issued separately: https://github.com/NCATSTranslator/ReasonerAPI/pull/176. 

I've kept this PR clean of the proposed 'type' change for `QEdge.relation` and `Edge.relation` but you may wish to seriously consider the proposal for TRAPI release 1.0 of enforcing CURIE values for these properties.  

I'm told that Vlado suggested that KPs might sometimes not be able to return a CURIE for the `relation `property, but I emphasize that the Biolink Model actually imposes that constraint.

TranslatorReasonerAPI.yaml
```diff
@@ -300,16 +325,27 @@ components:
...
+            as originally specified by, or curated by inference from, the
+            original external source of knowledge. Note that this should
+            often be specified as predicate ontology term CURIE, although
+            this may not be strictly enforced.
         subject:
           type: string
           example: https://omim.org/entry/603903
-          description: Corresponds to the id of subject node of this edge
+          description: >-
+            Corresponds to the map key CURIE of the
```
CaseyTa, line 339: I don't believe the subject and object need to be CURIEs in the query graph, they could just be identifiers internal to the message. See Patrick's example in his new PR (e.g., 'n0' and 'n1'): https://github.com/NCATSTranslator/ReasonerAPI/pull/175/files.

TranslatorReasonerAPI.yaml
```diff
@@ -208,12 +214,24 @@ components:
...
+        A instance of EdgeBinding is a single KnowledgeGraph Edge mapping,
+        identified by the corresponding 'id' object key identifier of the
+        Edge within the Knowledge Graph. Instances of EdgeBinding may include
+        extra annotation (such annotation is not yet fully standardized).
       properties:
         id:
           type: string
-          description: A knowledge-graph edge id, i.e. the `id` of a KEdge
+          description: The key identifier of a specific KnowledgeGraph Edge.
+        score:
```
CaseyTa, line 226: I think the `score` property should be introduced and discussed in a separate PR from the documentation PRs since it's a functional change to TRAPI.

The same comment may also apply for specifying `additionalProperties: True` . This change doesn't functionally change anything since that's the default behavior anyway, but this may give the impression that an explicit decision has been made when I'm not sure one has been made. I'll defer this to @edeutsch and @patrickkwang

CaseyTa (changes requested):

TranslatorReasonerAPI.yaml
patrickkwang: Agreed, adding this score property is a functional change, and should be independent of these documentation updates. `additionalProperties: true` is not a functional change since it is the default, and I think it's a good idea to make it explicit for readability. @edeutsch made the same addition in a separate PR.

RichardBruskiewich: Done.  See Commit  https://github.com/NCATSTranslator/ReasonerAPI/pull/170/commits/1c96bb08015ed286a0da1e5873a02e8d1fb9a216

CaseyTa (approved):

patrickkwang (approved):

RichardBruskiewich:
@CaseyTa, @patrickkwang, @edeutsch,

What's the SOP for merging pull requests that are now "approved" and free of conflicts with master.? Do one of you guys do this or do I have the power to do it?

edeutsch:
Patrick or I just merge it when it seems like there are enough approvals.

vdancik (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -245,13 +251,19 @@ components:
...
+        extra annotation (such annotation is not yet fully standardized).
       properties:
         id:
           type: string
-          description: A knowledge-graph edge id, i.e. the `id` of a KEdge
+          description: The key identifier of a specific KnowledgeGraph Edge.
       additionalProperties: true
       required:
         - id
+      additionalProperties: true
```
rtroper, line 266: Looks like this line duplicates line 263

RichardBruskiewich: > Looks like this line duplicates line 263

Yeah...  @rtroper ... not sure how that slipped in. Fixed and committed the patch.

RichardBruskiewich:
@edeutsch fixed a bug in the PR. ready again to merge

rtroper (approved):

Issue #143, opened 2020-08-06
change /predicates endpoint to /knowledge_map


vdancik:
In addition to providing node types and predicates, we should also provide information about the types of node identifiers that the ARA/KP can consume and produce.

edeutsch:
@vdancik there seems to be substantial interest in this. Do you want to put together a pull request with what you have in mind? I don't have a clear idea on what you're proposing. Perhaps something like the SRI NodeNormalizer's get_curie_prefixes endpoint?

https://nodenormalization-sri.renci.org/apidocs/#/Interfaces/get_get_curie_prefixes

vdancik:
Here is an example knowledge map for ChEMBL target information. In addition to specifying predicates, node types, and node identifiers, information about attributes is also included.

```
{
    "knowledge_map": {
        "nodes": {
            "biolink:ChemicalSubstance": {
                "id_prefixes": [
                    "ChEMBL:",
                    "inchikey:"
                ],
                "attributes": [
                    {
                        "type": "data:0990",
                        "source": "ChEMBL",
                        "names": [
                            "name"
                        ]
                    },
                    {
                        "type": "data:1001",
                        "source": "ChEMBL",
                        "names": [
                            "synonym"
                        ]
                    }
                ]
            },
            "biolink:Disease": {
                "id_prefixes": [
                    "MESH:",
                    "EFO:"
                ],
                "attributes": [
                    {
                        "type": "data:3668",
                        "source": "ChEMBL",
                        "names": [
                            "name", "synonym"
                        ]
                    }
                ]
            }
        },
        "predicates": [
            {
                "subject": "biolink:ChemicalSubstance",
                "predicate": "biolink:treats",
                "target": "biolink:Disease",
                "source": "ChEMBL",
                "relations": [
                    "treats"
                ],
                "attributes": [
                    {
                        "type": "OPMI:0000367",
                        "source": "ChEMBL",
                        "names": [
                            "max phase for indication"
                        ]
                    },
                    {
                        "type": "NCIT:C25641",
                        "source": "ChEMBL",
                        "names": [
                            "ClinicalTrials","DailyMed","ATC","FDA"
                        ]
                    },
                ]
            }
        ]
    }
}
```

southalln:
[...]

Issue #142, opened 2020-08-06
Add metadata endpoint


vdancik:
We should add a metadata endpoint, say /info, that would return information what server is and what it can do:

```
{
    "title": "best ARA ever",
    "description": "This is the best ARA ever.",
    "version": "1.1.0",
    "TRAPI": {
        "version": "1.0.0",
        "synchronous support": true,
        "asynchronous support": false,
        "callback support": false,
        "batch support": true,
        "extended": true,
        "supported operations": ["query"]
    }
}
```

vdancik:
Adding "supported operations" as suggested by @cbizon

newgene:
@vdancik a general comment is I would not include whitespace in the field names and replace it with `_` to make it one word.

Some general fields like `version` are also available from standard OpenAPI as `info.version` field, so you might have two places to put the same info. Alternative is to put additional TRAPI info within API metedata:

```
info:
  description: "This is the best ARA ever."
  title: "best ARA ever"
  version: "1.1.0"
  x-trapi:
           "TRAPI": "1.0.0",
           "synchronous support": true,
           "asynchronous support": false,
           "callback support": false,
           "extended": true,
           "supported operations": ["query"],
           "KGX export URL": null
```

But I can see an argument to have everything in a self-contained endpoint, so I am fine either way, just need to keep duplicated info in-sync.

patrickkwang:
This `x-trapi` OpenAPI extension would be a good way to include this sort TRAPI-specific, service-level information. It might could be a bit more concise if the fields "synchronous support", "asynchronous support", "callback support", and "batch support" are inferred from the presence or absence of the associated endpoints:
`/query/sync` (this is the current synchronous interface),
`/query/async/pull` (asynchronous interface with polling),
`/query/async/push` (asynchronous interface with callback),
and batch versions: `/batch/query/sync`, `/batch/query/async/pull`, `/batch/query/async/push`.

edeutsch:
addressed with x-trapi

PR #171, opened 2020-10-29
Metadata support

vdancik:
Addresses issues #143 and #142.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -639,9 +639,3 @@ components:
                 List of operations supported/implemented by this
                 knowledge source
               example: [query]
-            kgx_export_url:
-              type: string
-              description: >-
-                A link to data from this knowledge source exported in
-                kowledge-graph-export format
-              example: https://translator.ncats.io/kgx/bestARAever
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -564,12 +564,12 @@ components:
             values and indicates whether higher or lower values are more
             significant.
           properties:
-            from: 
+            from:
               type: string
               description: >-
                 Lower bound for allowed numeric values. Must be a valid float
                 number or +/-infinity.
-            to: 
+            to:
               type: string
               description: >-
                 Upper bound for allowed numeric values. Must be a valid float
@@ -589,7 +589,7 @@ components:
                     - smaller
                 more_specific:
                   description: >-
-                    Indicates whether larger or smaller values are more 
+                    Indicates whether larger or smaller values are more
                     specific.
                   type: string
                   enum:
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -29,6 +29,7 @@ tags:
 paths:
   /predicates:
     get:
+      deprecated: true
       tags:
         - predicates
       summary: Get supported relationships by source and target
@@ -53,6 +54,19 @@ paths:
                     'biolink:Gene':
                       - biolink:directly_interacts_with
                       - biolink:decreases_activity_of
+  /knowledge_map:
+    get:
+      tags:
+        - knowledge_map
+      summary: Knowledge-map representation of this knowledge source.
+      responses:
+        '200':
+          description: >-
+            Returns knowledge-map representation of this knowledge source.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/KnowledgeMap'
   /query:
     post:
       tags:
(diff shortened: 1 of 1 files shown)


TranslatorReasonerAPI.yaml
```diff
@@ -422,3 +449,199 @@ components:
       description: A Compact URI
       externalDocs:
         url: https://www.w3.org/TR/2010/NOTE-curie-20101216/
+    KnowledgeMap:
+      type: object
+      properties:
+        nodes:
+          type: object
+          description: >-
+            List of semantic types provided by this knowledge source.
```
patrickkwang, line 458: This is of type object rather than list.

TranslatorReasonerAPI.yaml
```diff
@@ -422,3 +449,199 @@ components:
...
+          additionalProperties:
```
patrickkwang, line 459: It's unclear what the keys are meant to be.

TranslatorReasonerAPI.yaml
```diff
@@ -422,3 +449,199 @@ components:
...
+            $ref: '#/components/schemas/KMNode'
+        predicates:
```
patrickkwang, line 461:
```suggestion
        edges:
```

TranslatorReasonerAPI.yaml
```diff
@@ -422,3 +449,199 @@ components:
...
+        attributes:
+          type: array
+          items:
+            $ref: '#/components/schemas/KMAttribute'
+          description: >-
+            Attributes associated with this node.
+      required:
+        - id_prefixes
+      additionalProperties: false
+    KMPredicate:
```
patrickkwang, line 504:
```suggestion
    KMEdge:
```
The "predicate" is a fairly well-defined string property of this meta-graph edge.

TranslatorReasonerAPI.yaml
```diff
@@ -422,3 +449,199 @@ components:
       description: A Compact URI
       externalDocs:
         url: https://www.w3.org/TR/2010/NOTE-curie-20101216/
+    KnowledgeMap:
```
patrickkwang, line 452: "MetaGraph"?

TranslatorReasonerAPI.yaml
```diff
@@ -53,6 +54,32 @@ paths:
...
+      summary: Knowledge-map representation of this knowledge source.
+      responses:
+        '200':
+          description: >-
+            Returns knowledge-map representation of this knowledge source.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/KnowledgeMap'
+  /info:
```
patrickkwang, line 70: I'm not convinced that we need this. I think that this information should be available in the service's OpenAPI schema itself.

TranslatorReasonerAPI.yaml
patrickkwang: This is very similar to the Service Provider's MetaKG.

edeutsch (changes requested):
Also Travis issues with line lengths > 79 and spaces at end of lines: https://travis-ci.com/github/NCATSTranslator/ReasonerAPI/builds/195308628

vdancik:
Actually, this should be milestone 1.1 and labeled "metadata" and not "documentation".

TranslatorReasonerAPI.yaml
```diff
@@ -422,3 +449,199 @@ components:
...
+          type: string
+          description: >-
+            Description od this knowledge source
+          example: This is the best ARA ever.
+        version:
+          type: string
+          description: >-
+            Version od this knowledge source
+          example: 1.0.1
+        TRAPI:
```
bill-baumgartner, line 595: For the TRAPI classes that have `additionalProperties: true` (Query, KnowledgeGraph, QueryGraph, QNode, and QEdge), would those additional properties be described in the metadata here as well?

TranslatorReasonerAPI.yaml
vdancik: Since we are talking about using knowledge map to represent metadata for KGX as well, maybe we should use categories/predicates rather then nodes/edges. I agree that, as it is now, nodes/predicates does not make sense.

TranslatorReasonerAPI.yaml
vdancik: This proposal only addresses the structure of the knowledge graph.

colleenXu:
Hi, a few thoughts given what I've been working on:

- for numeric range, "from" and "to" don't give info on whether those boundaries are inclusive or exclusive (accept and include the lower/higher bound or not). I was inspired by XML Schema/SHACL and was thinking of doing the following: (1) require one of the two properties representing a lower bound, either minInclusive or minExclusive, and (2) require one of the two properties representing a higher bound, either maxInclusive and maxExclusive. 

- numeric measures are not always related to confidence (they can be about the specificity of the relationship, and more) and directions are not always as simple as larger or smaller. I was therefore thinking of a required numeric_direction_info field for a "dictionary". The keys are then one or more directions ("closer to 0", "more negative", etc.) and the values are what those directions mean ("more evidence", "more specific"). 

Example of the direction issue: a measure could have negative and positive numbers, where...
      - being closer to 0 means no effect/association
      - a more negative number means a stronger association with a certain direction. An example from multiomics provider was a measure where more negative numbers meant a gene mutation led to more cell death after drug exposure (aka higher drug sensitivity)
      -  a more positive number means a stronger association with the opposite direction. For that example from multiomics provider, more positive numbers meant a gene mutation led to less cell death after drug exposure (aka higher drug resistance)

- not all edge properties represent numeric measures, and this schema seems more specific to numeric measures. Other edge properties could be related to provenance, publications or links related to the association, categorical measures, and context/relevance. I'm therefore thinking of describing multiple edge properties, that may have diff structures from each other, to accommodate this...

TranslatorReasonerAPI.yaml
vdancik: This proposal is accepted

TranslatorReasonerAPI.yaml
vdancik: We settled on node/edge terminology

vdancik:
@colleenXu we will revisit attribute metadata in a separate PR after reorganization of the Attribute schema (#186)

RichardBruskiewich:
Glancing at the latest metadata  schema (I give  `KMNode` here but similar comment  for  `KMEdge`):

```
KMNode:
      type: object
      description: >-
        Description of semantic types provided by this knowledge source.
      properties:
        id_prefixes:
          type: array
          description: >-
            List of CURIE prefixes that this knowledge source understands and
            accepts on the input.
          items:
            type: string
          example: ['ChEMBL:', 'inchikey:']
        count:
          type: integer
          description: >-
            Total number of node instances known to this knowledge source,
            -1 if the number is not known
        count_by_source:
          type: object
          description: >-
            Number of node instances by their source
          additionalProperties:
            type: integer
          example:
            ChEMBL: 1234
      required:
        - id_prefixes
 ```
The descriptions  use the terminology '**knowledge source**' which _seems_ to refer  to the Translator component (ARA or  KP)  exposing the TRAPI  interface, but then there is  secondary usage of the  word  '**source**' (i.e.  `count_by_source`)  which _seems_ to refer to  the backend  (non-Translator,  external,  3rd party)  knowledge  sources  wrapped by the  KP  (or whose knowledge is published after analysis, by an ARA).  

I would   personally  reserve the nomenclature  `knowledge source` for the  external 3rd party  databases _et al._ and decide to  use  another term for  the  TRAPI   component  itself.  Perhaps  not use the word 'provided' either,  but  rather,  a   term  like  'exported'  or  'exposed'.  For  example, the KMNode description could  perhaps  rather  read:

```
Description of semantic types exposed by this TRAPI web service.
```

TomConlin:
Consider some linkage / pointer to where one may find *what* this "TRAPI web service" understands a curie to be for.

marcdubybroad (approved):

TranslatorReasonerAPI.yaml
vdancik: Agreed, /info was removed and relevant information will be provided by x-translator and x-trapi tags

vdancik:
Replaced by PR #215

PR #175, opened 2020-11-05
Add simple Message example

patrickkwang:


edeutsch (changes requested):
Seems okay, except can we make n0 specifically Ebola? The QueryGraph as written is too vague to be answerable. If n0 is specifically MONDO:0005737, the example makes a whole lot more sense.

edeutsch (approved):
looks good, thanks!

bill-baumgartner (approved):

CaseyTa (approved):

vdancik (approved):

PR #176, opened 2020-11-05
(Q)Edge.relation should be an ontology CURIE

RichardBruskiewich:
If TRAPI and the Biolink Model are to remain closely aligned at the fundamental semantic level, then the `Edge.relation` property of a KnowledgeGraph, and probably the QEdge of the QueryGraph, absolutely need to have a value of CURIE dereferencing a formal external 3rd party ontology class definition. 

I believe this to be a release 1.0 'bug' in the TRAPI schema.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -410,8 +410,8 @@ components:
         predicate:
           $ref: '#/components/schemas/BiolinkRelation'
         relation:
-          type: string
-          example: upregulates
+          $ref: '#/components/schemas/CURIE'
+          example: RO:0002447
           description: Lower-level relationship type of this edge
         subject:
           $ref: '#/components/schemas/CURIE'
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -332,8 +332,8 @@ components:
               items:
                 $ref: '#/components/schemas/BiolinkRelation'
         relation:
-          type: string
-          example: upregulates
+          $ref: '#/components/schemas/CURIE'
+          example: RO:0002447
           description: Lower-level relationship type of this edge
         subject:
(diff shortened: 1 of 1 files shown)


TomConlin:
+1

TomConlin:
Has there been any discussion of  categories  for edges?  
"lower level type of an edge" is getting pretty close to a meaning of category

RichardBruskiewich:
> 
> 
> Has there been any discussion of categories for edges?
> "lower level type of an edge" is getting pretty close to a meaning of category

Actually, edges have `predicate` (formerly `edge_label`) already. I think the TRAPI comment about 'lower level' is a bit misleading. All that it meant by this is the "original (predicate) relationship coming from the (external, non-Translator) knowledge source" for that 'relationship edge'

Thus the `predicate` slot already has the analogous role for relationship edges that `category` slot has for concept nodes.

TomConlin:
ah so lower level in this case is approximately  'original'

vdancik:
I believe that this should stay as currently documented - use of CURIEs for relations should be encouraged, but not formally required.

vdancik (changes requested):
I would keep `type: string` and the following description:

"The relationship type term of this edge, originally specified by, or curated by inference from, the original source of knowledge. This should generally be specified as predicate ontology CURIE."

CaseyTa:
> I would keep `type: string` and the following description:
> 
> "The relationship type term of this edge, originally specified by, or curated by inference from, the original source of knowledge. This should generally be specified as predicate ontology CURIE."

And perhaps also change the example to a CURIE

ehinderer (approved):

edeutsch:
(Q)Edge.relation was previously removed as not relevant, so this PR is not moot and is being withdrawn after discussion on 2021-07-22

PR #177, opened 2020-11-06
Add provenance holder for Edge

patrickkwang:
This information does not belong in attributes, but additionalProperties are explicitly disallowed.

I expect this to be more fully specified in future releases, but for now, I need somewhere to put provenance information.


edeutsch (commented):
Very bold! I agree that piling all this in attributes is not a great idea. There has been conversation in the background with Richard and others about this. I suggest we tackle this head-on on the Nov 12 call

patrickkwang:
Consensus is that this should go in `attributes` for now.

PR #178, opened 2020-11-11
Updating node_bindings, NodeBinding, and Node descriptions

ehinderer:
I copied the description that @RichardBruskiewich logged for EdgeBinding and edge_bindings from #170 and changed it for Nodes as necessary. Those descriptions fit my understanding of it. As for Node itself, I'm not sure what more to add, since it's a fairly generic concept.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -245,8 +245,8 @@ components:
           description: >-
             An instance of NodeBinding is a single KnowledgeGraph Node mapping,
             identified by the corresponding 'id' object key identifier of the
-            Node within the Knowledge Graph. Instances of NodeBinding may 
-            include extra annotation (such annotation is not yet fully 
+            Node within the Knowledge Graph. Instances of NodeBinding may
+            include extra annotation (such annotation is not yet fully
             standardized).
       required:
         - id
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -221,7 +221,7 @@ components:
             bindings where the dictionary keys are the key identifiers of the
             Query Graph nodes and the associated values of those keys are
             instances of NodeBinding schema type (see below). This value is an
-            array of NodeBindings since a given query edge may have multiple
+            array of NodeBindings since a given query node may have multiple
             knowledge graph Node bindings in the result.
           additionalProperties:
             type: array
@@ -362,7 +362,8 @@ components:
       type: object
       description: >-
         A node in the knowledge graph which represents some biomedical
-        concept.
+        concept. Nodes are identified by the keys in the KnowldgeGraph 
+        Node mapping.
       properties:
         name:
           type: string
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -361,8 +361,8 @@ components:
     Node:
       type: object
       description: >-
-        A node in the knowledge graph which represents some biomedical
-        concept. Nodes are identified by the keys in the KnowldgeGraph 
+        A node in the KnowledgeGraph which represents some biomedical
+        concept. Nodes are identified by the keys in the KnowledgeGraph
         Node mapping.
       properties:
         name:
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -216,7 +216,13 @@ components:
       properties:
         node_bindings:
           type: object
-          description: List of QNode-KNode bindings.
+          description: >-
+            The dictionary of Input Query Graph to Result Knowledge Graph node
+            bindings where the dictionary keys are the key identifiers of the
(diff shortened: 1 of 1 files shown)


edeutsch (changes requested):
Seems pretty good, but automatic checking yielded line too long:
TranslatorReasonerAPI.yaml
  248:80    error    line too long (81 > 79 characters)  (line-length)
Would you fix that?

edeutsch (changes requested):
Travis is fussy:
  248:74    error    trailing spaces  (trailing-spaces)

  249:71    error    trailing spaces  (trailing-spaces)

ehinderer:
True! I wish there was some kind of syntax highlighting on GitHub.

edeutsch (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -349,7 +360,9 @@ components:
         - object
     Node:
       type: object
-      description: A node in the knowledge graph
+      description: >-
+        A node in the knowledge graph which represents some biomedical
+        concept.
```
CaseyTa, line 365: It may be helpful to add in the description that the nodes are identified by the keys in the KnowldgeGraph Node mapping.

TranslatorReasonerAPI.yaml
```diff
@@ -216,7 +216,13 @@ components:
       properties:
         node_bindings:
           type: object
-          description: List of QNode-KNode bindings.
+          description: >-
+            The dictionary of Input Query Graph to Result Knowledge Graph node
+            bindings where the dictionary keys are the key identifiers of the
+            Query Graph nodes and the associated values of those keys are
+            instances of NodeBinding schema type (see below). This value is an
+            array of NodeBindings since a given query edge may have multiple
```
CaseyTa, line 224: I think "a given query edge" should be "a given query node"

CaseyTa (changes requested):

CaseyTa:
A question popped into my head while reviewing: since the result binding specifies the bound KG edge, and since the KG edge points to specific subject and object nodes, then are the KG node bindings necessary or are they redundant?

patrickkwang:
>A question popped into my head while reviewing: since the result binding specifies the bound KG edge, and since the KG edge points to specific subject and object nodes, then are the KG node bindings necessary or are they redundant?

This is very insightful. Reducing the payload size as well as preventing certain internal inconsistencies both sound nice. There is just one issue I foresee. Some services, when given a qedge with a symmetric predicate (e.g. related_to) or no predicate, will bind a kedge that goes the opposite direction from the qedge. This would make inferring the node bindings difficult or impossible for some types of query graphs.

TranslatorReasonerAPI.yaml
ehinderer: Yes, this is a typo that I accidentally carried over from Richard's version.

TranslatorReasonerAPI.yaml
ehinderer: I don't know what this is referring to. I don't see mention of a node mapping in the KnowledgeGraph object here.

ehinderer:
> A question popped into my head while reviewing: since the result binding specifies the bound KG edge, and since the KG edge points to specific subject and object nodes, then are the KG node bindings necessary or are they redundant?

I was under the impression that one of the purposes of having node and edge bindings in the result was so that ARAs could add some layer of evaluation on top of the answers. For instance, the Explanatory Agent adds explanations to the results section which point to edge bindings (and could also point to node bindings). It tells the user things like "here is why we recommend this particular association that was pulled from the KP, which relates to this part original question you posed."

TranslatorReasonerAPI.yaml
CaseyTa: I'm referring to the `nodes` property of `KnowledgeGraph`. Someone looking at the definition of `Node` may wonder where the CURIE should be specified. I think it would be helpful to mention that the CURIE is the key in the `nodes` dictionary / map.

ehinderer: I was confused because `nodes` definition says that it's a "list of nodes in the KnowledgeGraph." Are you saying that this is a list of CURIE strings, or a dictionary of CURIE string: Node? If so, should I go ahead and change the definition of `node ` in this PR?

vdancik (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -349,7 +360,10 @@ components:
         - object
     Node:
       type: object
-      description: A node in the knowledge graph
+      description: >-
+        A node in the knowledge graph which represents some biomedical
+        concept. Nodes are identified by the keys in the KnowldgeGraph
```
rtroper, line 365: Small typo here: KnowldgeGraph

ehinderer:
@edeutsch @CaseyTa Is this okay to be merged or is there still something needing addressed?

PR #179, opened 2020-11-12
Add explicit additionalProperties

edeutsch:


patrickkwang (approved):

bill-baumgartner (commented):
Thanks for making these explicit. I think that helps, especially for those like myself who are still learning the OpenAPI spec. My question is, should all of these be `true`?  Are there use cases for additional properties on `NodeBinding` instances, for example? Or is it simply common practice to not lock down the ability to add additional properties?

edeutsch:
> Thanks for making these explicit. I think that helps, especially for those like myself who are still learning the OpenAPI spec. My question is, should all of these be `true`? Are there use cases for additional properties on `NodeBinding` instances, for example? Or is it simply common practice to not lock down the ability to add additional properties?

This PR is just about making explicit that which was already implicit.

If we want to change any of these, that should be a subsequent PR after this one is merged.

In general, I would advocate leaving it open unless there is a compelling reason why we want to make sure there are no additional properties are specified. This is often done in cases where we cannot require certain properties because of alternate use cases and there have been historical cases of implementers making up their own property names for the same concepts.

bill-baumgartner (approved):
Thanks for that explanation @edeutsch. Makes sense.

CaseyTa (approved):

PR #181, opened 2020-11-17
'score' property  added to EdgeBinding

RichardBruskiewich:


vdancik (changes requested):
Not necessary as a separate property, can be provided as an edge attribute.

edeutsch:
We might want to add this in the future, but let's see what the EPC group comes up with

ehinderer (approved):

vgardner-renci:
https://github.com/NCATSTranslator/ReasonerAPI/pull/247 accomplished this

Issue #185, opened 2020-11-23
Can the Attributes schema model be generalised to reflect division of semantic concerns thus richer knowledge representation?


RichardBruskiewich:
The current ReasonerAPI (1.0) OpenAPI schema for the **Attribute** model in TRAPI is the following:

```yaml
    Attribute:
      type: object
      description: >-
        Generic attribute for a node or an edge that expands key-value pair
        concept by including a type of this attribute from a suitable ontology,
        a source of this attribute, and (optionally) a url with additional
        information about this attribute.
      properties:
        name:
          type: string
          description: >-
            Human-readable name or label for the attribute. If appropriate,
            should be the name of the semantic type term.
          example: PubMed Identifier
        value:
          example: 32529952
          description: >-
            Value of the attribute. May be any data type, including a list.
        type:
          $ref: '#/components/schemas/CURIE'
          description: >-
            CURIE of the semantic type of the attribute. For properties
            defined by the Biolink model this should be a biolink CURIE,
            otherwise, if possible, from the EDAM ontology. If a suitable
            identifier does not exist, enter a descriptive phrase here and
            submit the new type for consideration by the appropriate
            authority.
          example: EDAM:data_1187
        url:
          type: string
          description: >-
            Human-consumable URL to link out and provide additional information
            about the attribute (not the node or the edge).
          example: https://pubmed.ncbi.nlm.nih.gov/32529952
        source:
          type: string
          description: Source of the attribute, preferably as a CURIE prefix.
          example: UniProtKB
      required:
        - type
        - value
      additionalProperties: false
```

The limitation with this data model is that it doesn't adequately document context of interpretation for an attribute, when such context may be necessary and helpful. At best, it conflates such context with the nature of the attribute value.  This deficient is especially notable in attempts to represent so-called "evidence, provenance, confidence/context" information for TRAPI subgraphs being returned.

The key problem is that all of the fields, especially `type`, only refer to the value of the `attribute`. The `type` is expected to be a bioinformatics data type (i.e. semantics as typically qualified by the EDAM ontology). For example, a 'p value' is of data type `http://edamontology.org/data_1669`; however, the only interpretation of the model is that this statistic somehow a score related to the `Edge` but doesn't speak of its context of interpretation nor origin. It may be claimed that the statistic is somehow self descriptive but is that really true?

Here we will suggest modifications to the Attribute schema then we'll discuss the benefits.

[...]

PR #186, opened 2020-12-02
Attribute schema object revisions to enhance semantics

RichardBruskiewich:
Attribute schema object revisions to enhance semantics, especially, for basic reporting of evidence, provenance and confidence metadata, in response to TRAPI issue https://github.com/NCATSTranslator/ReasonerAPI/issues/185 and elaborated further in issue https://github.com/NCATSTranslator/ReasonerAPI/issues/192.

The proposed new model clearly distinguishes between the **data type** of the `value `of an attribute (`value_type` and `value_type_name`) _versus_ the attribute itself, as defined by `attribute_type` and `attribute_name`, giving the **context of its usage, interpretation or role in the system** (e.g. using an identifier of a person playing the role as the direct source for evidence, as a data curator of the record, or as an author in a publication, etc.).

This revision is proposed based on the conviction that a `value` and its `value_type` ("data type") - without the additional context of usage - Biolink, SEPIO, or otherwise - for the given attribute - cannot always be unambiguously interpreted, even with EDAM ontology overloading of the semantics of the original value `type` field.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -474,27 +474,27 @@ components:
         (i.e. using an identifier of a person as a source for evidence,
         as a data curator,as a publication author, etc.)
       properties:
-        attribute_type:
+        attribute_type_id:
           $ref: '#/components/schemas/CURIE'
           description: >-
             CURIE of the ontology term defining the attribute. For properties
-            defined by the Biolink model this should be a biolink CURIE.
+            defined by the Biolink model this should be a Biolink CURIE.
             For evidence provenance confidence properties defined by the
-            SEPIO model, this should be a SEPIO CURIE.
+            SEPIO model, this should generally be a SEPIO CURIE.
           example: SEPIO:0000130
-        attribute_name:
+        attribute_from_source:
           type: string
           description: >-
-            Human-readable name or label for the attribute. If appropriate,
-            should be the name of the ontology term of the attribute type but
-            could be any string which expresses the intent of the attribute.
+            Attribute term as defined by the source defining the attribute.
+            The data type is 'string' but the contents of the field could
+            also be a CURIE of a third party ontology term.
           example: Assertion Authored By
           nullable: true
         value:
           example: 32529952
           description: >-
             Value of the attribute. May be any data type, including a list.
-        value_type:
+        value_type_id:
           $ref: '#/components/schemas/CURIE'
           description: >-
             CURIE of the semantic type of the attribute value. ,
@@ -503,26 +503,35 @@ components:
             submit the new type for consideration by the appropriate
             authority.
           example: EDAM:data_1187
-        value_type_name:
+        value_type_from_source:
+          $ref: '#/components/schemas/CURIE'
+          description: >-
+            Semantic type of the attribute value as defined by the source
+            defining the attribute. The data type is 'string' but the contents
+            of the field could also be a CURIE of a third party ontology term.
+          example: p value
+          nullable: true
+        value_source:
           type: string
           description: >-
-            Human-readable name or label for the value type, usually the name
-            of the value type semantic type term.
-          example: PubMed Identifier
-        url:
+            Source of the attribute value, preferably as a CURIE namespace.
+          example: UniProtKB
+          nullable: true
+        value_url:
           type: string
           description: >-
             Human-consumable URL to link out and provide additional information
             about the attribute value (not the node or the edge).
           example: https://pubmed.ncbi.nlm.nih.gov/32529952
           nullable: true
-        source:
+        description:
           type: string
-          description: Source of the attribute, preferably as a CURIE prefix.
-          example: UniProtKB
+          description: >-
+            Human-readable description for the attribute and its value.
+          example: Assertion Authored By Dr. Trans L. Ator
           nullable: true
       required:
-        - attribute_type
+        - attribute_type_id
         - value
         - value_type
       additionalProperties: false
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.1
 info:
   description: OpenAPI for NCATS Biomedical Translator Reasoners
-  version: 1.0.0-beta
+  version: 1.1.0-beta
   title: OpenAPI for NCATS Biomedical Translator Reasoners
   contact:
     email: edeutsch@systemsbiology.org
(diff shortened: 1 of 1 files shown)


RichardBruskiewich:
Further discussion on issue  https://github.com/NCATSTranslator/ReasonerAPI/issues/192, but  proposed revisions not yet incorporated.

edeutsch (approved):

vdancik (approved):

CaseyTa (approved):

rtroper (approved):

edeutsch:
My understand is that after the discussion with Chris Mungall, Matt Brush, Patrick, me, the change from the previous pull request is this:
```
Previous PR #186                New proposal
-------------------------       ------------------------------
attribute_type_id               attribute_type_id
attribute_from_source        *  original_attribute_name
value                           value
value_type_id                   value_type_id
value_url                       value_url
description                  *  value_description
value_type_from_source       *  REMOVE
value_source                 *  attribute_source
```

Please re-review, okay to merge?

TranslatorReasonerAPI.yaml
```diff
@@ -470,46 +470,75 @@ components:
...
+          description: >-
+            The 'key' of the attribute object, holding a CURIE of an ontology
+            property defining the attribute (preferably the CURIE of a
+            Biolink association slot). This property captures the relationship
+            asserted to hold between the value of the attribute, and the node
+            or edge from  which it hangs. For example, that a value of
+            '0.000153' represents a p-value supporting an edge, or that
+            a value of 'ChEMBL' represents the original source of the knowledge
+            expressed in the edge.
+          example: Biolink:has_p-value_evidence, Biolink:has_original_source
```
patrickkwang, line 491: These do not appear to be proper biolink CURIEs.

TranslatorReasonerAPI.yaml
```diff
@@ -470,46 +470,75 @@ components:
...
+        value_type_id:
           $ref: '#/components/schemas/CURIE'
           description: >-
-            CURIE of the semantic type of the attribute. For properties
-            defined by the Biolink model this should be a biolink CURIE,
-            otherwise, if possible, from the EDAM ontology. If a suitable
-            identifier does not exist, enter a descriptive phrase here and
-            submit the new type for consideration by the appropriate
-            authority.
+            CURIE describing the semantic type of an  attribute's value. Use
```
patrickkwang, line 512: Is the `value_type_id` always, sometimes, or never implied by the `attribute_type_id`? Maybe some examples of the two used together would be helpful.

TranslatorReasonerAPI.yaml
```diff
@@ -470,46 +470,75 @@ components:
...
+          example: UniProtKB
+          nullable: true
+        value_url:
           type: string
           description: >-
-            Human-consumable URL to link out and provide additional information
-            about the attribute (not the node or the edge).
+            Human-consumable URL linking to a web document that provides
+            additional information about an  attribute's value (not the node
+            or the edge fom which it hangs).
```
patrickkwang, line 531:
```suggestion
            or the edge from which it hangs).
```

patrickkwang (changes requested):

TranslatorReasonerAPI.yaml
ehinderer: I also cannot find these examples in the biolink data model. Are they perhaps ones that have not made it into the documentation?

TranslatorReasonerAPI.yaml
```diff
@@ -470,46 +470,75 @@ components:
...
-            otherwise, if possible, from the EDAM ontology. If a suitable
-            identifier does not exist, enter a descriptive phrase here and
-            submit the new type for consideration by the appropriate
-            authority.
+            CURIE describing the semantic type of an  attribute's value. Use
+            a Biolink class if possible, otherwise a term from an external
+            ontology. If a suitable CURIE/identifier does not exist, enter a
+            descriptive phrase here and submit the new type for consideration
+            by the appropriate authority.
           example: EDAM:data_1187
```
ehinderer, line 517:
```suggestion
          example: EDAM:data_1669
```

ehinderer: Stay consistent with previous examples? This one is for "p-value" in EDAM

TranslatorReasonerAPI.yaml
ehinderer: The attribute_type_id seems to describe the semantic association between the value and the object that contains that value, whereas the value_type_id describes only the value itself. I think the similarity would depend on the specificity of the attribute_type_id. Something specific like has_p_value_evidence will probably only ever link to a p-value-related term. Whereas something like has_original_source could be a publication, an assay, another knowledge source, or something else.

TranslatorReasonerAPI.yaml
```diff
@@ -470,46 +470,75 @@ components:
...
+            or the edge fom which it hangs).
           example: https://pubmed.ncbi.nlm.nih.gov/32529952
           nullable: true
-        source:
+        description:
           type: string
-          description: Source of the attribute, preferably as a CURIE prefix.
-          example: UniProtKB
+          description: >-
+            Human-readable description for the attribute and its value.
```
ehinderer, line 537:
```suggestion
            Human-readable description of the source which provides the attribute and its 
            value.
```

TranslatorReasonerAPI.yaml
```diff
@@ -470,46 +470,75 @@ components:
...
           description: >-
-            Human-consumable URL to link out and provide additional information
-            about the attribute (not the node or the edge).
+            Human-consumable URL linking to a web document that provides
+            additional information about an  attribute's value (not the node
+            or the edge fom which it hangs).
           example: https://pubmed.ncbi.nlm.nih.gov/32529952
           nullable: true
-        source:
+        description:
```
ehinderer, line 534: Is there a conflict with shadowing names from outer scope? Consider renaming to value_url_description if so.

ehinderer (changes requested):
I've proposed a couple minor changes and echo Patrick's concerns about some of the example CURIEs not seeming valid.

TranslatorReasonerAPI.yaml
mbrush: Yes, these are examples of edge properties proposed to support capture of EPC in Attributes, but are not yet in Biolink.  Vetting and getting these into Biolink is our next step to support use of the Attribute model for consistent and computable representation of edge metadata.

TranslatorReasonerAPI.yaml
mbrush: This is exactly right Eugene.  The idea here is to allow for very granular/specific properties that can be tied to a specific value type. We are exploring if/when this is practical and useful - using the examples in the attribute_types inventory.  If we end up supporting properties specific to each value type, then this field becomes redundant (in that the value type can always be looked up in the Biolink Model).  

But  it may be that there is value in duplicating this info in the message itself to avoid the need to lookup form Biolink .  And we anticipate that there may be cases where a KP wants a value_type that is more precise than we want in Biolink (e.g. indicating that a curie for a Publication represents a DOI vs PMID).  We will assess this as we work through real use cases/examples.

TranslatorReasonerAPI.yaml
mbrush: The way that github shows changes here is a bit confusing.  The 'description' field on line 534 in green here is meant to hold a description of the info captured by the attribute. Nothing to do with the value_url.   The fork in Richard B's repo that this PR is based on may provide a clearer view of the current proposal (absent of github's interpretation of line changes): https://github.com/STARInformatics/ReasonerAPI/blob/issue-185-attribute-schema-revisions_R1.1/TranslatorReasonerAPI.yaml#L470

TranslatorReasonerAPI.yaml
mbrush: See comment above about looking to Richard B's fork . . .

TranslatorReasonerAPI.yaml
patrickkwang: If these slots are coming to biolink, that's fine, but we've been using the prefix (lowercase) "biolink". Is that correct?

TranslatorReasonerAPI.yaml
```diff
@@ -470,46 +470,75 @@ components:
...
+            a Biolink class if possible, otherwise a term from an external
+            ontology. If a suitable CURIE/identifier does not exist, enter a
+            descriptive phrase here and submit the new type for consideration
+            by the appropriate authority.
           example: EDAM:data_1187
-        url:
+        attribute_source:
+          type: string
+          description: >-
+            The source of the core assertion madeby the key-value pair of an
```
patrickkwang, line 521:
```suggestion
            The source of the core assertion made by the key-value pair of an
```

PR #187, opened 2020-12-03
Additional TRAPI 1.0 documentation

RichardBruskiewich:


edeutsch (approved):

bill-baumgartner (approved):

vdancik (approved):

rtroper (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -295,18 +295,26 @@ components:
     KnowledgeGraph:
       type: object
       description: >-
-        A thought graph associated with this result. This will commonly be a
-        linear path subgraph from one concept to another, but related items
-        aside of the path may be included.
+        A thought graph associated with this result. This will commonly be
```
patrickkwang, line 298: I think we discussed avoiding this "thought graph" terminology. Would it be out-of-line to use "knowledge graph" in the description here? Also, the use of "result" here is a bit weird because the Message contains a list of "results". How about: "The knowledge graph associated with a set of results."

TranslatorReasonerAPI.yaml
```diff
@@ -295,18 +295,26 @@ components:
...
+        a linear path subgraph from one concept to another, but related items
```
patrickkwang, line 299: My impression is that for most questions, the knowledge graph is not linear, even when the query graph is. For one-hop questions, for instance, it's a star, with each Result corresponding to one point of the star.

TranslatorReasonerAPI.yaml
RichardBruskiewich: sure. np.

TranslatorReasonerAPI.yaml
RichardBruskiewich: Sure. You are more in the know about  this.  I'll modify this  accordingly  and  re-commit.

patrickkwang (approved):

edeutsch (approved):

PR #194, opened 2021-01-08
Prepare for 1.0.0 final

patrickkwang:


edeutsch (approved):
Yeah, I saw these pesky warnings as well. Seems like a fine fix. I'm also seeing warnings for these, and so they probably also need the same fix:
Query.message
Response.message
NodeBinding.id
Attribute.type
Edge.subject
Edge.object

patrickkwang:
The ones I fixed were functional problems because the `$ref` actually overrode the `nullable` (and the `type`). All these others are just `description` and `example` fields, so I'm less concerned about them. The solution would be just to remove the overridden fields or move them into the referenced schemas.

edeutsch:
okay, fair enough. fine to proceed AFAIC

CaseyTa (approved):

edeutsch (approved):

vdancik (approved):

PR #200, opened 2021-01-26
Strawman operations{} component

edeutsch:
Add a flexible container that can hold a list of actions, a map of options, and then potentially message URIs or even multiple messages


jzollars:
@kadelakun

dkoslicki:
@edeutsch @patrickkwang Is this on the TRAPI agenda to get into the "official" TRAPI spec?

edeutsch:
yes, we'll get this into TRAPI 1.2 before release. should be easy.

edeutsch:
This PR is withdrawn in favor of #262

PR #203, opened 2021-01-28
Strawman table of Attribute types for discussion

edeutsch:


edeutsch:
This is a bit more easily viewed at https://github.com/NCATSTranslator/ReasonerAPI/blob/edeutsch-AttributeTypes/AttributeExamples.md

rtroper:
For our KPs, it'll be important to be able to specify cutoffs using < and > rather than exact values (e.g. for p-value or weights) such as can be done for a Smart API query. Do we anticipate this will be supported at the query level in TRAPI? Or is this kind of filtering something that ARAs will do after getting a set of results back?

marcdubybroad (commented):
I'd separate p_value and probability, adding 'biolink:probability' to the biolink edge properties list if possible; the higher probability will indicate stronger association, vs lower for p_value.
Great idea to have standadized type table that we can draw from.

edeutsch:
@rtroper, I  agree I think your comment is more relevant to another topic, not this one, which is just about reporting, not querying. Your questions is somewhat related to #105, but I don't think that's it either. So, really, you ought to register your own issue I think. In some way, it is related to the operations and workflow effort. A specific example would help.

RichardBruskiewich:
TRAPI  meeting  decision:  @vdancik and @RichardBruskiewich to  set up a  Google Doc to capture inventory  of these terms.

RichardBruskiewich:
> TRAPI meeting decision: @vdancik and @RichardBruskiewich to set up a Google Doc to capture inventory of these terms.

[Google Doc Inventory is  here](https://docs.google.com/spreadsheets/d/1-ilDWePMLniA9Tha5J6HHHFylO5w6lZ9FFfA2Mp5oro)

vdancik:
PR #259 provides mechanism for declaring attributes in `/meta_knowledge_graph`

edeutsch:
This was useful to get the discussion going, but the EPC working group has done substantial additional work to address this topic, and this PR is obsolete. Being withdrawn after discussion at the call 2021-07-22

PR #208, opened 2021-02-11
Alter location of statement of TRAPI support

edeutsch:
I think this is the proposal that was on the table? Use x-trapi extension to signal which version of TRAPI is supported by a service. Service is free to use the top-level version to specify the version of their service.


patrickkwang (approved):
I think this is exactly right.

vdancik (approved):

brettasmi (approved):

CaseyTa (approved):

rtroper (approved):

PR #215, opened 2021-02-18
Metadata support

vdancik:
This serves as replacement for /predicates endpoint. Addresses issue #143 and replaces PR #171.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -64,7 +64,7 @@ paths:
       responses:
         '200':
           description: >-
-            Returns meta knowledge graph representation of this TRAPI web 
+            Returns meta knowledge graph representation of this TRAPI web
             service.
           content:
             application/json:
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -623,7 +623,7 @@ components:
             service understands and accepts on the input.
           items:
             type: string
-          example: ['ChEMBL:', 'inchikey:']
+          example: ['CHEMBL.COMPOUND', 'INCHIKEY']
         count:
           type: integer
           description: >-
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -56,19 +56,19 @@ paths:
                     'biolink:Gene':
                       - biolink:directly_interacts_with
                       - biolink:decreases_activity_of
-  /knowledge_map:
+  /meta_knowledge_graph:
     get:
       tags:
-        - knowledge_map
-      summary: Knowledge-map representation of this TRAPI web service.
+        - meta_knowledge_graph
+      summary: Meta knowledge graph representation of this TRAPI web service.
       responses:
         '200':
           description: >-
-            Returns knowledge-map representation of this TRAPI web service.
+            Returns meta knowledge graph representation of this TRAPI web service.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/KnowledgeMap'
+                $ref: '#/components/schemas/MetaKnowledgeGraph'
   /query:
     post:
       tags:
@@ -593,7 +593,7 @@ components:
         URI.
       externalDocs:
         url: https://www.w3.org/TR/2010/NOTE-curie-20101216/
-    KnowledgeMap:
+    MetaKnowledgeGraph:
       type: object
       description: >-
         Knowledge-map representation of this TRAPI web service.
@@ -643,7 +643,7 @@ components:
     KMEdge:
       type: object
       description: >-
-        Edge in a knowledge map describing relationship between a subject
+        Edge in a meta knowledge map describing relationship between a subject
         Biolink class and an object Biolink class.
       properties:
         subject:
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -605,15 +605,15 @@ components:
             Collection of node categories provided by this TRAPI web service,
             indexed by Biolink class CURIEs.
           additionalProperties:
-            $ref: '#/components/schemas/KMNode'
+            $ref: '#/components/schemas/MetaNode'
         edges:
           type: array
           description: >-
             List of edges/predicates provided by this TRAPI web service.
           items:
-            $ref: '#/components/schemas/KMEdge'
+            $ref: '#/components/schemas/MetaEdge'
           minItems: 1
-    KMNode:
+    MetaNode:
       type: object
       description: >-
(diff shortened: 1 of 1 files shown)


TranslatorReasonerAPI.yaml
```diff
@@ -55,6 +56,19 @@ paths:
                     'biolink:Gene':
                       - biolink:directly_interacts_with
                       - biolink:decreases_activity_of
+  /knowledge_map:
```
patrickkwang, line 59: Re: naming, I'd prefer meta_graph or meta_knowledge_graph. That would align with the SmartAPI/BTE notion of a MetaKG. It's not obvious to me that the phrase "knowledge map" would be interpreted by a naive reader much differently from "knowledge graph".

TranslatorReasonerAPI.yaml
```diff
@@ -579,3 +593,97 @@ components:
...
+      properties:
+        id_prefixes:
+          type: array
+          description: >-
+            List of CURIE prefixes for the node category that this TRAPI web
+            service understands and accepts on the input.
+          items:
+            type: string
+          example: ['ChEMBL:', 'inchikey:']
+        count:
```
patrickkwang, line 627: I'm not sure that/how the Ranking Agent would use count information if it were provided, so I'm inclined to put it in the extended schema.

TranslatorReasonerAPI.yaml
```diff
@@ -579,3 +593,97 @@ components:
...
+        Description of a node category provided by this TRAPI web service.
+      properties:
+        id_prefixes:
+          type: array
+          description: >-
+            List of CURIE prefixes for the node category that this TRAPI web
+            service understands and accepts on the input.
+          items:
+            type: string
+          example: ['ChEMBL:', 'inchikey:']
```
patrickkwang, line 626: We should align all of the examples with the biolink context prefixes: https://github.com/biolink/biolink-model/blob/master/context.jsonld

In this case, we may want something like `['CHEMBL.COMPOUND', 'INCHIKEY']`.

Also, should these prefixes include the trailing colon? My feeling is no.

edeutsch: I agree, examples should show official curie prefixes without colons

TranslatorReasonerAPI.yaml
edeutsch: The results on the last vote on this are in the minutes from Dec 10. They were quite split although /knowledge_map was the winner with a slight majority. I would be fine with reconsidering and revoting on the endpoint name as we complete this for real.

TranslatorReasonerAPI.yaml
kevinxin90: Wondering for id_prefixes, does the order an id prefix appears in the list indicate priority?

vdancik: There is no intent to indicate preference, service promises to handle any of the prefixes.

TranslatorReasonerAPI.yaml
vdancik: done

TranslatorReasonerAPI.yaml
```diff
@@ -579,3 +594,70 @@ components:
...
+      type: object
+      description: >-
+        Description of a node category provided by this TRAPI web service.
+      properties:
+        id_prefixes:
+          type: array
+          description: >-
+            List of CURIE prefixes for the node category that this TRAPI web
+            service understands and accepts on the input.
+          items:
```
patrickkwang, line 625: Should we insist that this array be non-empty?

TranslatorReasonerAPI.yaml
```diff
@@ -579,3 +594,70 @@ components:
...
+          additionalProperties:
+            $ref: '#/components/schemas/KMNode'
+        edges:
+          type: array
+          description: >-
+            List of edges/predicates provided by this TRAPI web service.
+          items:
+            $ref: '#/components/schemas/KMEdge'
+    KMNode:
+      type: object
```
patrickkwang, line 616: These values are objects with exactly one property, which begs the question, should the values just be the id-prefix arrays, directly?

vdancik: I would say no, we will also have attributes here as well

TranslatorReasonerAPI.yaml
vdancik: Makes sense, actually the array of edges in the meta knowledge graph should be non-empty as well.

TranslatorReasonerAPI.yaml
```diff
@@ -579,3 +594,70 @@ components:
...
+      description: >-
+        Description of a node category provided by this TRAPI web service.
+      properties:
+        id_prefixes:
+          type: array
+          description: >-
+            List of CURIE prefixes for the node category that this TRAPI web
+            service understands and accepts on the input.
+          items:
+            type: string
```
patrickkwang, line 626:
```suggestion
            type: string
          minItems: 1
```

TranslatorReasonerAPI.yaml
```diff
@@ -579,3 +594,70 @@ components:
...
+            Collection of node categories provided by this TRAPI web service,
+            indexed by Biolink class CURIEs.
+          additionalProperties:
+            $ref: '#/components/schemas/KMNode'
+        edges:
+          type: array
+          description: >-
+            List of edges/predicates provided by this TRAPI web service.
+          items:
+            $ref: '#/components/schemas/KMEdge'
```
patrickkwang, line 614:
```suggestion
            $ref: '#/components/schemas/KMEdge'
          minItems: 1
```

TranslatorReasonerAPI.yaml
```diff
@@ -579,3 +594,72 @@ components:
...
+          minItems: 1
+    KMNode:
```
patrickkwang, line 616: We should probably rename these to MKNode and MKEdge, or just MNode and MEdge.

vdancik: Or maybe MetaNode, MetaEdge to complement MetaKnowledgeGraph

patrickkwang (approved):

edeutsch (approved):

CaseyTa (approved):

rtroper (approved):

sstemann:
@NCATSTranslator/trapi-wg added reviewers, per the 3/11 TRAPI WG call, the WG is looking for review/approval/dissent

bill-baumgartner (approved):

marcdubybroad (approved):

brettasmi (approved):
👍

Lilly-25:
Hello,
Can someone guide me on how to download the knowledge graph here?

Issue #216, opened 2021-03-09
x-translator and SmartAPI


edeutsch:
After today's Architecture call that demoed some new features at SmartAPI, I check in on my APIs at SmartAPI and I see this:
![image](https://user-images.githubusercontent.com/12707718/110527241-90251d80-80cb-11eb-8548-25d0bc0f50e0.png)

Not good, I suppose. Clicking on the validate link, I see:
![image](https://user-images.githubusercontent.com/12707718/110527318-aaf79200-80cb-11eb-901a-d7c958f3c0fc.png)

Our yaml looks like this:
```
  x-translator:
    component: ARA
    team: Expander Agent
```
and the official TRAPI schema yaml says:
```
  x-translator:
    component: INSERT-EITHER-ARA-OR-KP-HERE
    team: INSERT-TRANSLATOR-TEAM-NAME-HERE
```

So, the question is: why am I seeing this error? I think I am following the official schema.

Ideas?

colleenXu:
You should be compliant if you make team a one-element array, with the element being "Expander Agent".

The official schema that SmartAPI uses is [here](https://github.com/NCATSTranslator/translator_extensions/blob/production/x-translator/smartapi_x-translator_schema.json). Chris Bizon recently requested that teams be an array only (see the closed issues); Chunlei and I reviewed it, brought it up in the Architecture meeting last week, and then approved the change. 

I believe the "TRAPI schema yaml" is somehow not in sync with the JSON schema that SmartAPI is using.

PR #217, opened 2021-03-11
add current x-translator, x-trapi requirements

colleenXu:
note reasoner tag is being deprecated. Added on request of Eric, see issue #216

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -10,14 +10,23 @@ info:
     url: http://www.apache.org/licenses/LICENSE-2.0.html
   termsOfService: INSERT-URL-HERE
   x-translator:
-    component: INSERT-EITHER-ARA-OR-KP-HERE
-    team: INSERT-TRANSLATOR-TEAM-NAME-HERE
+    component: INSERT-EITHER-ARA-OR-KP-OR-ARS-OR-Utility-HERE
+    team:
+      - INSERT-ONE-TRANSLATOR-TEAM-NAME-PER-ELEMENT-HERE
     externalDocs:
       description: >-
         The values for component and team are restricted according to this
-        external JSON schema
+        external JSON schema. See schema and examples at url
       url: https://github.com/NCATSTranslator/translator_extensions/blob/\
-        main/x-translator/smartapi_x-translator_schema.json
+        production/x-translator/
+  x-trapi:
+    version: INSERT-TRAPI-VERSION-HERE
+    externalDocs:
+      description: >-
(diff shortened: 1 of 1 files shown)


vdancik (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -10,14 +10,23 @@ info:
     url: http://www.apache.org/licenses/LICENSE-2.0.html
   termsOfService: INSERT-URL-HERE
   x-translator:
-    component: INSERT-EITHER-ARA-OR-KP-HERE
-    team: INSERT-TRANSLATOR-TEAM-NAME-HERE
+    component: INSERT-EITHER-ARA-OR-KP-OR-ARS-OR-Utility-HERE
```
brettasmi, line 13: This is a little confusing with 'EITHER' and then a list of three things. Also, the capitalization `Utility` isn't consistent with the rest of the ALL CAPS.

brettasmi (approved):
Looks good! I made one minor comment on formatting.

rtroper (approved):

patrickkwang (approved):

TranslatorReasonerAPI.yaml
colleenXu: I put EITHER to show that this was an enum (have to pick one of the values). This can be removed if the enum is clear.

"Utility" is a current value in the enum. Changing this involves changing the schema, and then asking the 4 or so APIs that have this value to change it...

edeutsch (approved):

TranslatorReasonerAPI.yaml
brettasmi: For the either issue, I was pointing out that that word typically suggests two choices, though of course it can mean more. Changing to "ONE-OF" might make it clearer, but I don't really care, especially with all the other approvals.

Hear you on Utility 👍 

Thanks Colleen!

colleenXu: Okay, we'll see if others want changes as well.

I hope this PR is helpful for registering APIs. I didn't know that TRAPI and translator_extension schemas weren't quite aligned, and that this was causing initial errors with registry API validation...

marcdubybroad (approved):

ehinderer (approved):

bill-baumgartner (approved):

vdancik:
Should this go to v1.1.x branch rather then master branch?

edeutsch:
probably, but they will merge together in just over a week and this is not a schema change, so probably doesn't matter IMO.

PR #221, opened 2021-03-23
Update fields that should be changed to be more obvious

edeutsch:
Effort to try to avoid all the new registrations that look like:
image


bill-baumgartner (approved):
Had the exact same thought.

vdancik (approved):

rtroper (approved):

patrickkwang (approved):

ehinderer (approved):

PR #222, opened 2021-03-24
Alternate proposal for constraints

edeutsch:
Here is the alternate proposal for what QNode and QEdge constraints could look like as discussed last week.
This is an alternative PR to #157

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -425,6 +425,15 @@ components:
             with is_set=True is connected to a QNode (n2) with is_set=True,
             each n1 must be connected to at least one n2.
           default: false
+        constraints:
+          type: array
+          description: >-
+            A list of contraints applied to a query node.
+            If there are multiple items, they must all be true (equivalent
+            to AND)
+          items:
+            $ref: '#/components/schemas/QueryConstraint'
+          nullable: true
       additionalProperties: true
     QEdge:
       type: object
@@ -466,6 +475,15 @@ components:
             Corresponds to the map key identifier of the
             object concept node anchoring the query filter
             pattern for the query relationship edge.
+        constraints:
+          type: array
+          description: >-
+            A list of contraints applied to a query edge.
+            If there are multiple items, they must all be true (equivalent
+            to AND)
+          items:
+            $ref: '#/components/schemas/QueryConstraint'
+          nullable: true
       additionalProperties: true
       required:
         - subject
@@ -682,3 +700,50 @@ components:
         - predicate
         - object
       additionalProperties: false
+    QueryConstraint:
+      type: object
+      description: >-
+        Generic query constraint for a query node or query edge
+      properties:
+        name:
+          type: string
+          description: >-
+            Human-readable name or label for the constraint. If appropriate,
+            it should be the name of the semantic type term.
+          example: molecular mass
+          nullable: false
+        id:
+          $ref: '#/components/schemas/CURIE'
+          description: >-
+            CURIE of the concept being constrained. For properties
+            defined by the Biolink model this should be a biolink CURIE,
+            otherwise, if possible, from the EDAM ontology. If a suitable
+            identifier does not exist, enter a descriptive phrase here and
+            submit the new type for consideration by the appropriate
+            authority.
+          example: EDAM:data_0844
+          nullable: false
+        not:
+          type: boolean
+          default: false
+        operator:
+          type: string
+          enum:
+            - ==
+            - '>'
+            - <
+            - 'in'
+            - 'matches'
+          default: ==
+        value:
+          example: 57.0
+          description: >-
+            Value of the attribute. May be any data type, including a list.
+            If the value is a list and there are multiple items, at least one
+            must all be true (equivalent to OR)
+      required:
+        - name
+        - id
+        - operator
+        - value
+    additionalProperties: false
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -754,10 +754,11 @@ components:
           description: >-
             Value of the attribute. May be any data type, including a list.
             If the value is a list and there are multiple items, at least one
-            comparison must be true (equivalent to OR). If 'value' is a data type
-            object, the keys of the object MAY be treated as a list. A 'list' data
-            type paired with the '>' or '<' operators will encode extraneous
-            comparisons, but this is permitted as it is in SQL and other languages.
+            comparison must be true (equivalent to OR). If 'value' is of data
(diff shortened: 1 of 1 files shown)


edeutsch:
Here is an example that is approximately equivalent to https://github.com/NCATSTranslator/ReasonerAPI/pull/157#issuecomment-790241941:
```
query_graph:
  nodes:
    n0:
      categories:
      - biolink:Gene
      constraints:
      - name: Chromosome band
        id: NCIT:C13432
        operator: matches
        negated: false
        value: 11q13.*
      - name: gene_family
        id: NCIT:C26004
        operator: in
        negated: false
        value:
        - BBSome
        - Antisense RNAs
      - name: gene_family
        id: NCIT:C26004
        operator: in
        negated: true
        value:
        - defensin
    n1:
      categories:
      - biolink:Disease
      constraints:
      - name: Congenital Abnormality
        id: NCIT:C2849
        operator: ==
        value: true
    n2:
      categories:
      - biolink:ChemicalSubstance
      constraints:
      - name: molecular mass
        id: EDAM:data_0844
        operator: >
        value: 57.0
      - name: molecular mass
        id: EDAM:data_0844
        operator: <=
        value: 45.0
```

vdancik (approved):

CaseyTa:
I like this structure better than #157 for the reasons already mentioned by Eric in his [comment ](https://github.com/NCATSTranslator/ReasonerAPI/pull/157#issuecomment-806211446) on #157. 

I also think having the `id` will be important for specificity, but I assume `id` could also be added to #157 if people prefer that syntax. One question though, since `id` is required, I assume people would prefer to operate on the `id` rather than the `name`, so should `name` still be required? I think it's still nice to have in the query for human readability, but perhaps `name` could be nullable / not required. 

For the example constraint on Congenital Abnormality, the intent seems to be to constrain the diseases to diseases that are congenital abnormalities, but given the operator `==`, I would strictly interpret this example query to mean nodes that are exactly NCIT:C2849 (not sure if others agree). I think this constraint could also be written using an additional node and edge (n1 biolink:Disease) -- biolink:subclass_of -- (n3 id=NCIT:C2849). Should both formats be allowed for the same constraint? Is one format preferred over the other? My opinion is that using biolink:subclass_of would be clearer. 

Numerical constraints will need a field for units. 

Are we still targeting the 1.1 release for these constraints? I think there are still too many open questions for people to get this implemented by the 1.1 implementation deadline.

TranslatorReasonerAPI.yaml
```diff
@@ -682,3 +700,51 @@ components:
...
+          type: boolean
+          default: false
+        operator:
+          type: string
+          enum:
+            - ==
+            - '>'
+            - <
+            - in
+            - matches
```
brettasmi, line 737: I think that this probably requires a description field to explain the various conditions and how they might behave given different `value` types.

brettasmi:
I agree with all of @CaseyTa's points and interpretation of the example using the `==` operator as exactly matching that CURIE.

Using `subclass_of`, that would be my preferred expression of the concept, though I will say that it would be nice to know if that's really a one-hop to the next subclass or n-hops to any subclass of 'congenital abnormality.' We ran into this recently doing some case studies with metabolic diseases.

I am generally of the opinion that it would be good to get this into the spec and iterate to improve it. That said, it seemed like the consensus is that this isn't going to be used much anytime soon anyway, so I wouldn't mind pushing it to 1.2 if there are still strong opinions or ambiguities that need resolving.

edeutsch:
I have fixed this up a little based on discussion above. Some things to finalize at the meeting below. Feel free to provide comments here prior to the meeting:
1. Call the property 'not' or 'negated'?
2. Should the Constraint.name be optional or required?
3. == SHOULD NOT be used in the sense of an "is a" subclass relationship?
4. Added unit_id and unit_name. Do we want both? They are optional but both are recommended for quantities taking units
5. Do we really want to squeeze this into TRAPI 1.1 at the last minute (we decided YES last week)?
6. Are the added descriptions sufficiently detailed and encoding the desired behavior?
7. Merge it?

CaseyTa:
> Do we really want to squeeze this into TRAPI 1.1 at the last minute (we decided YES last week)?

Apologies, I was on vacation last week. I did not intend to re-open something that had already been decided.

edeutsch:
This is my take on the alignment between Attribute and Constraint:
```
Attribute                 Constraint
attribute_type_id        ~ operator   (attribute_type_id and operator are not dissimilar is some ways)
attribute_source         x            (source is the client)
original_attribute_name  x            (not relevant)
value                    = value      (exact match)
value_type_id            = id         (match the properties names?)
  ?                      ~ name       (name is similar to the 'value_type_from_source' that was recently discarded from Attribute)
value_url                x            (not relevant)
description              x            (probably not really relevant)
  ?                      + not        (probably not useful in Attribute since this is captured in attribute_type_id. Or maybe useful?)
  ?                      + unit_id    (this was entertained in Attribute, but deferred for 1.2? Reconsider and align?)
  ?                      + unit_name  (this was entertained in Attribute, but deferred for 1.2? Reconsider and align?)
```

do we make changes after pondering this?

bill-baumgartner:
I generally like this proposal over #157 as well, in particular for the use of `id` as @CaseyTa  pointed out above.

> 1. Call the property 'not' or 'negated'?

My vote is for `negated` but would be fine with `not`.

> 2. Should the Constraint.name be optional or required?

My vote is for `optional`.

> 3. == SHOULD NOT be used in the sense of an "is a" subclass relationship?

I initially interpreted the `== NCIT:C2849` constraint above similarly to @CaseyTa and @brettasmi. The intent seems to be to constrain the node to a single class (NCIT:C2849) while excluding its subclasses. This constraint is acting on the class of the node while all of the other constraints in the example above are acting on attributes of the node. Is that a fair statement? For clarity it might make sense to separate these kinds of constraints. One option could be to add a flag to indicate that the subClassOf hierarchy shouldn't be traversed. So, in the example above, `n1` could be modeled as the following:

```
    n1:
      id:
      - NCIT:C2849
      direct_class_only: true
```

where the `direct_class_only` flag (note: I'm sure there is a better name for this flag) when set to `true` would constrain the node to be the explicitly listed id(s). The default value would be `false` indicating that the node should match the listed id(s) and any subclasses.

stevencox (approved):

marcdubybroad (approved):

vdancik (approved):

CaseyTa (approved):

bill-baumgartner (approved):

brettasmi (approved):

rtroper (approved):

yakaboskic (approved):

PR #225, opened 2021-04-01
Add Change Log

edeutsch:


CaseyTa (approved):

bill-baumgartner (approved):

brettasmi (approved):
This looks good.

I'd suggest that going forward we start this right away and keep a running list of changes with links to the relevant PRs as we merge them in. It'd be nice to be able to click a link to see the discussion related to each change. For example: 

- QNode.id pluralized to ids and changed from CURIE or array of CURIEs to just array of CURIEs with array length > 0 #202 

I don't think it's worth doing for 1.1 since it would be a lot of work to collect all of those, but it would be easier to do piece by piece going forward.

rtroper (approved):

vdancik (approved):

ehinderer (approved):

PR #240, opened 2021-04-14
Make QueryConstraint.operator optional

### LINKED ISSUE #239: QueryConstraint.operator is required but also has a default value
As far as I can tell, this is not actually prohibited by OpenAPI 3.0, but it is confusing and causes problems with pydantic.

patrickkwang:
Fixes #239.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -820,6 +820,5 @@ components:
       required:
         - name
         - id
-        - operator
         - value
       additionalProperties: false


edeutsch (approved):

ehinderer:
Looks to me like the problem lies with "==" being a default value of operator, not with operator being required. I can't think of an example of a constraint where an operator isn't used, that doesn't make sense to me. Along the same lines, if someone is defining a constraint, they probably know what operator they want to use. I'm not sure it's necessary to default to "==".

TranslatorReasonerAPI.yaml
```diff
@@ -820,6 +820,5 @@ components:
       required:
         - name
         - id
-        - operator
```
ehinderer, line 823: Why remove operator requirement from constraint? I can't think of a valid constraint that does not contain an operator. 

If the driving factor was that operator had a default value, per #239, consider removing default value from operator? A constraint definition should be unambiguous anyway.

ehinderer (changes requested):

patrickkwang:
@ehinderer I agree, removing the default value is a better solution.

PR #242, opened 2021-04-15
Disallow additional properties in QNode and QEdge

patrickkwang:


vdancik (approved):

CaseyTa (approved):

brettasmi:
What was the original anticipated use for these additionalProperties? Is anyone using them now?

marcdubybroad (approved):

ehinderer:
I also don't understand what the original intention was for additional properties on the QEdge and QNode, or why it's being proposed for removal. Sorry if this was answered in one of the meetings already. 

FYI if you are intending on disallowing these, you may also look at lines 392 and 401 where additionalProperties are referenced to QNode and QEdge, respectively. Not sure if that'll need to be removed as well.

edeutsch:
I think this comes out of discussion that was had last week. I was on vacation and was not there, so I can't really say.

By closing off additionalProperties I am a bit concerned that we are making experimental/extended features more difficult.
For example, in ARAX we have some experimental properties like:
QEdge.exclude (boolean) If set to true, then all subgraphs containing this edge are excluded from the final results

You can use this if you want to explicitly want to disallow a connection between two nodes, for example, if there is a "biolink:contraindicated_for" edge in the KG.
Like: Which drugs are related to disease X except contraindicated ones.
You'd have two edges, one "biolink:associated_with" and one "biolink:contraindicated_for" (excluded=true)
By setting additionalProperties=true, we can no longer legally do this. We can add it to our (ARAX) schema, of course (which we have), but it won't pass the validator any more I assume. So as long as we don't do validation, it would be okay.

I started this whole journey with "How should a server respond if it doesn't fully understand the question". And it seems that the discussion last week ended up with "don't let clients say anything valid that shouldn't be understandable". Which I think closes off a lot of creativity. And/or doesn't really ultimately address the problem.

It seems like there are two approaches:
1) Let there be runtime errors if things don't seem to be correct
2) Try to anticipate everything that could be wrong and make sure it's a validation error

Since very few groups if any (we're not) are doing strict validation I suspect, 2 doesn't really solve the problem anyway.

rtroper (commented):
I'm interested to know the compelling reasons to remove. Otherwise, I'd be in favor of leaving it as-is, for now.

patrickkwang:
Consensus is that there are reasons to allow additional properties, and we will just make sure to document the expected server behavior upon encountering unexpected non-standard properties.

Issue #234, opened 2021-04-08
How to handle 'relation' in query edge


vdancik:
Right now (version 1.1.0-beta) the predicate is an array while the relation is a string in query edge. Should the relation:
1. stay a scalar
1. be an array
1. be removed

edeutsch:
Closing because solution 3 was implemented.

PR #243, opened 2021-05-05
Remove relation from query edge

vdancik:
This addresses inconsistency pointed out in #234

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -464,16 +464,6 @@ components:
             $ref: '#/components/schemas/BiolinkPredicate'
           minItems: 1
           nullable: true
-        relation:
-          type: string
-          example: RO:0002447
-          description: >-
-            Query constraint against the relationship type term of this edge,
-            as originally specified by, or curated by inference from, the
-            original external source of knowledge. Note that this should
-            often be specified as predicate ontology term CURIE, although
-            this may not be strictly enforced.
-          nullable: true
         subject:
           type: string
           example: https://omim.org/entry/603903
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -21,7 +21,7 @@ info:
       url: https://github.com/NCATSTranslator/translator_extensions/blob/\
         production/x-translator/
   x-trapi:
-    version: 1.1.0-beta
+    version: 1.2
     externalDocs:
       description: >-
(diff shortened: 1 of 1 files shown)


RichardBruskiewich:
Hi @vdancik and @edeutsch, are we now documenting the "source" relation of an edge somewhere else, in EPC attributes, in such a manner that it is accessible to some user query? 

I'd sort of like @sierra-moxon, @cmungall and @mbrush  to assess this proposal, since as far as I know (and my understanding of the Biolink Model maybe out-of-date at this moment), the `relation` field in the model is meant to record the _original_ predicate (e.g. RO term?) that an external knowledge source uses to characterize the relationship captured by the edge.

Whether or not we remove it from the `QEdge` would depend on whether or not there remains any use case for Translator users searching for edges in knowledge sources which have such an external predicate (e.g. RO term) in the knowledge graph. 

Perhaps not... But all the same, I'd prefer that @sierra-moxon, @cmungall and, perhaps, @mbrush, express their opinions about this before I approve this PR.

edeutsch:
Thanks Richard.

To clarify slightly, this PR only addresses the removal of the `relation` property in the *query edge* QEdge.

The troubles are:
1. predicates is now a list
2. but relation is still a scalar
3. if we switched relation to relations (a list), it's not clear how it should mesh with the predicates list
4. We have not observed anyone using relation since predicates (which must be a biolink model predicate) seems sufficient
For these reasons, we are proposing to remove it.

We actually have a similar problem with relation in the knowledge graph Edge, but that is not addressed in the PR.

vdancik:
I am all for removing relations from knowledge-graph edges and using attributes with `biolink:relation` type to provide that information. Constrained queries can then be used to filter results for a giver relation.

CaseyTa (approved):

cmungall:
relation is a holdover from previous translator iteration where we needed generic and specific predicates. now we just use predicate

RichardBruskiewich (approved):
Chris Mungall's comments helped me finalize my decision to approve. You can probably remove the other reviewers (including Chris.. he didn't explicitly do the review but his comment can be taken as an approval).

sierra-moxon (approved):

vdancik:
I have updated this PR to remove relation from knowledge graph edge to reflect a proposed change in the Biolink model [#761 deprecate relation, add "original_" subject, predicate, object edge properties](https://github.com/biolink/biolink-model/issues/761)

PR #245, opened 2021-05-14
Make value_type_id nullable as intended

edeutsch:
The released schema does not explicitly allow value_type_id to be nullable although this was intended by not making is required.


patrickkwang (approved):
Looks great!

vdancik (approved):

Issue #246, opened 2021-05-18
"Upgrading" a scoring attribute from Extended to Core schema


cartmanbeck:
In order to make it clear that scoring is an important (I would say necessary) part of the results that come back at every level, I would like to make the attribute for score part of the core schema. If a name for the attribute is needed, I would suggest something like "certainty_score", to make it clear that the default metric for a score should be how certain the tool is that they've answered the query sufficiently.

edeutsch:
Addressed in #247

PR #247, opened 2021-05-20
upgrade confidence and score from extended

edeutsch:
These already exist in the extended schema. They are already used by ARAX and the ARAX viewer as shown for a long time.
This is a proposal to make them a standard part of the core (and become recommended for all results?)
Refinement or alternatives are welcome.
Addresses #246

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -319,24 +319,10 @@ components:
           type: number
           format: float
           example: 163.233
-          description: Any numerical score associated with this result
-        score_name:
-          type: string
-          example: Jaccard distance
-          description: Name for the score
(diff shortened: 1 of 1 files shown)


CaseyTa:
I think a range should be specified somewhere, whether it be in each component's TRAPI yaml (if we expect a component's scoring range to be static) or somewhere within the TRAPI message so the client can more easily interpret the score rather than only using it for ranking.

edeutsch:
I have updated this PR. Please add your review based on the current state

patrickkwang (approved):

CaseyTa (approved):

rtroper (approved):

vdancik (approved):

PR #253, opened 2021-05-26
State explictly that results list SHOULD NOT be considered ordered

edeutsch:
Do we push this into TRAPI 1.1.2?
or push into the v1.2 branch?

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -219,8 +219,8 @@ components:
         results:
           description: >-
             List of all returned Result objects for the query posed.
-            The list SHOULD NOT be assumed to be ordered. The client SHOULD
-            reorder the results based on the 'score' property if it is present.
+            The list SHOULD NOT be assumed to be ordered. The 'score' property,
+             if present, MAY be used to infer result rankings.
           type: array
           items:
             $ref: '#/components/schemas/Result'
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -21,7 +21,7 @@ info:
       url: https://github.com/NCATSTranslator/translator_extensions/blob/\
         production/x-translator/
   x-trapi:
-    version: 1.1.1
+    version: 1.1.3
     externalDocs:
       description: >-
         The values for version are restricted according to the regex in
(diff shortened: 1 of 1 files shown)


CaseyTa (approved):

edeutsch:
I have updated the PR based on discussion on the call. Please add your reviews

TranslatorReasonerAPI.yaml
```diff
@@ -218,7 +218,9 @@ components:
       properties:
         results:
           description: >-
-            List of all returned Result objects for the query posed
+            List of all returned Result objects for the query posed.
+            The list SHOULD NOT be assumed to be ordered. The client SHOULD
+            reorder the results based on the 'score' property if it is present.
```
patrickkwang, line 223: Why do we need this last sentence? Do we want the client to do something other than not assuming the order matters?

edeutsch: The second sentence is intended to make it clear where the client SHOULD look to find the ordering intended by the server.

Rather than telling the client what they should do, would it be preferable to replace the last sentence with:

The intended order of the results SHOULD be obtained by sorting the 'score' property if it is present.

patrickkwang: I'd prefer something even less imperative, like:

> The `score` property, if present, can be used to infer result rankings.

edeutsch: okay, I agree that is better. I have updated the PR. How is the latest now?

patrickkwang (approved):

CaseyTa (approved):

TranslatorReasonerAPI.yaml
suihuang-ISB: If the ARA does not rank (i.e. the list is randomly ordered), will this be explicitly declared, or it can just return the same score  value for all results, e.g. all zeros. Redundancy?

edeutsch: The `score` property is not required by the schema. So therefore, if there is no `score` provided (or it is NULL, which is equivalent), then there is no ranking, since the list is declared in the documentation to be unordered,

marcdubybroad (approved):

vdancik (approved):

PR #254, opened 2021-05-27
Specificity of /meta_knowledge_graph

cbizon:
Based on the 5/25 Architecture call, the prevailing opinion was that the meta knowledge graph endpoint should return a minimal specific set of catetgories/predicates and rely on ARAs to understand the inference rules and make sense of them.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -685,9 +685,9 @@ components:
         nodes:
           type: object
           description: >-
-            Collection of the most specific node categories provided by 
-            this TRAPI web service, indexed by Biolink class CURIEs.   
-            A node category is only exposed here if there is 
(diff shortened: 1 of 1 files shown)


edeutsch (changes requested):
This change seems pretty well stated to me, but checks are failing because of line length limit overruns and other formatting issues.

cbizon:
OK, thanks for pointing that out @edeutsch, formatting is passing now.

edeutsch (approved):

bill-baumgartner (approved):

newgene (approved):

andrewsu (approved):

rtroper (approved):

marcdubybroad (approved):

CaseyTa (approved):

vdancik (approved):

PR #255, opened 2021-06-08
Proposal for TRAPI 1.2 asyncquery

edeutsch:
New proposal based on discussion 2021-06-03

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -22,6 +22,7 @@ info:
         production/x-translator/
   x-trapi:
     version: 1.2.0
+    asyncquery: REPLACE-WITH-true-IF-/asyncquery-IS-IMPLEMENTED-ELSE-false
     externalDocs:
       description: >-
         The values for version are restricted according to the regex in
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -229,7 +229,9 @@ components:
           format: uri
           pattern: ^https?://
           description: >-
-            Upon completion, the Response will be POSTed to this URL.
+            Upon completion, this server will send a POST request to the callback
+            URL with `Content-Type: application/json` header and request body
+            containing a JSON-encoded `Response` object.
         message:
           $ref: '#/components/schemas/Message'
           description: >-
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -229,9 +229,9 @@ components:
           format: uri
           pattern: ^https?://
           description: >-
-            Upon completion, this server will send a POST request to the callback
-            URL with `Content-Type: application/json` header and request body
-            containing a JSON-encoded `Response` object.
+            Upon completion, this server will send a POST request to the
+            callback URL with `Content-Type: application/json` header and
+            request body containing a JSON-encoded `Response` object.
         message:
           $ref: '#/components/schemas/Message'
           description: >-
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -42,9 +42,14 @@ tags:
       description: Documentation for the reasoner predicates function
       url: INSERT-URL-HERE-OR-REMOVE-EXTERNALDOCS-IF-NA
   - name: query
-    description: Query reasoner using a predefined question type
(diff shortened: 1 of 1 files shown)


marcdubybroad (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -167,6 +217,36 @@ components:
...
+        it requires a callback property.
+      x-body-name: request_body
+      type: object
+      properties:
+        callback:
+          type: string
+          format: uri
+          pattern: ^https?://
+          description: >-
+            Upon completion, the Response will be POSTed to this URL.
```
patrickkwang, line 232:
```suggestion
            Upon completion, this server will send a POST request to the callback
            URL with `Content-Type: application/json` header and request body
            containing a JSON-encoded `Response` object.
```

patrickkwang (changes requested):

newgene:
@edeutsch do we want to support "polling" style asyncquery? The advantage is the client-side can actively check for the status.

patrickkwang:
Can the server POST to the callback URL multiple times? Say if the server generates some results quickly and further results more slowly?

edeutsch:
> @edeutsch do we want to support "polling" style asyncquery? The advantage is the client-side can actively check for the status.

I originally thought polling style would be nicer because it gives the client more control, but in our discussion June 3, we arrived at the conclusion that callback was much simpler, stateless, and what ARS already wants and has implemented. And so for the moment we have decided to go callback only. If there is a groundswell to reverse that decision, let it be known!

edeutsch:
> 
> 
> Can the server POST to the callback URL multiple times? Say if the server generates some results quickly and further results more slowly?

Although this was discussed, I do not think we came to a stated conclusion on this. It is my recollection that @MarkDWilliams said that ARS already allows unlimited calls to the callback URL, true? We should definitely clarify in the docs whether this is explicitly desired and permitted. If yes, could perhaps specify a different Response.status depending on whether the results are interim results or final results.

patrickkwang (approved):

newgene:
@edeutsch if it's just for ARS purpose, I guess that's fine. But if my TRAPI client is not a web application, just a Python script, I don't really have the existence of a callback API at the client side (unless specifically implement one). This is the case if an ARA or a KP needs to consume another TRAPI endpoint.

An async query is not goanna be stateless anyway, it's matter of whether you want to implement "sessions" at the TRAPI server-side or its client-side. If we make TRAPI server state-less, we are making every client to implement "sessions". I would rather to have it at the server-side.

Having said that, if we decided to go with callback for now to just make ARS work, and worry about "polling" support later, I am fine with it.

edeutsch:
> @edeutsch if it's just for ARS purpose, I guess that's fine. But if my TRAPI client is not a web application, just a Python script, I don't really have the existence of a callback API at the client side (unless specifically implement one). This is the case if an ARA or a KP needs to consume another TRAPI endpoint.

Agreed. I presented this argument at the June 3 call, but it did not carry the day. For ARS and ARAs I think callback will be fine and easier for everyone. For an arbitrary Python script behind a NAT firewall, this is hard. (if there's no NAT, then it seems easy, just set up a listener)

> An async query is not goanna be stateless anyway, it's matter of whether you want to implement "sessions" at the TRAPI server-side or its client-side. If we make TRAPI server state-less, we are making every client to implement "sessions". I would rather to have it at the server-side.

I am not well versed in these things, but it seems to me that a request with a callback qualifies as stateless, while a polling scenario does not. Maybe I'm confused about what stateless means.

> Having said that, if we decided to go with callback for now to just make ARS work, and worry about "polling" support later, I am fine with it.

I think this is the path we are on, exactly.

edeutsch:
@MarkDWilliams do you have a comment on the comment above about *multiple* POSTs to the callback URL? interim results? final results? Already supported? Planned? not?

vdancik (approved):

CaseyTa (approved):

CaseyTa:
The x-trapi option already implies that `/asyncquery` isn't required to be implemented, but should we document explicitly that it's not required? Could be in the yaml or in the readme or policies markdowns.

bill-baumgartner (approved):

rtroper (approved):

yakaboskic (approved):

kshefchek (approved):

PR #256, opened 2021-06-14
Batch, Op&Wf: message -> messages[ ]

edeutsch:
TRAPI 1.1 already supports batching lite via ids[ ]
But more extensive batching seems necessary in 1.2
Presumably at its simplest this is just switching message to messages[ ]
Op&Wf group also wants multiple messages[ ] for merge and diff operations
Seems simple
But there are huge implications for existing code that is designed to accept and emit a single message.
Other options that preserve some backwards compatibility for single-message communication is possible.
What shall we do?

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -209,7 +209,7 @@ components:
           nullable: true
       additionalProperties: true
       required:
-        - message
+        - messages
     Message:
       description: >-
         The message object holds the main content of a Query or a Response in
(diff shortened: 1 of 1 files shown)


CaseyTa:
The schema changes seem fine. 

Would all providers be expected to implement support for messages with more than 1 message or can a KP specify that it only supports a single message item? 

If there's no upper limit to the number of messages, KPs which previously responded quickly enough to not need to implement `/asyncquery` may now need to implement `/asyncquery`.

vgardner-renci:
During TRAPI WG call on 6/17/2021 it was decided that the level of batching we have now is sufficient for our purposes.

Issue #257, opened 2021-06-14
Advertise node and edge properties in /meta_knowledge_graph


cbizon:
TRAPI now lets a query use constraints to return only nodes or edges that have particular attribute values.

But how are callers to know what are the allowed properties to constrain?

They could (should) be advertised in the /meta_knowledge_graph endpoint, presumably also with some type info.

edeutsch:
Vote here:
https://github.com/NCATSTranslator/ReasonerAPI/discussions/251#discussioncomment-841949
to adjust the priority of this item

edeutsch:
Being addressed in #259

PR #259, opened 2021-06-17
meta_knowledge_graph attributes

vdancik:
addresses #257

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -706,6 +706,12 @@ components:
             type: string
           minItems: 1
           example: [CHEMBL.COMPOUND, INCHIKEY]
+        attributes:
+          type: array
+          description: >-
+            Node attributes provided by this TRAPI web service.
+          items:
+            $ref: '#/components/schemas/MetaAttribute'
       required:
         - id_prefixes
       additionalProperties: false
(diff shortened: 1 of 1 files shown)


TranslatorReasonerAPI.yaml
```diff
@@ -737,11 +743,51 @@ components:
...
+          description: >-
+            Source of an attribute provided by this TRAPI web service.
+          example: infores:chembl
+        original_attribute_names:
+          type: array
+          items:
+            type: string
+          description: >-
+            Names of an the attribute as provided by the source.
+        costraint_use:
```
CaseyTa, line 777: typo: `constraint_use`

TranslatorReasonerAPI.yaml
```diff
@@ -737,11 +743,51 @@ components:
...
+          type: boolean
+          description: >-
+            Indicates whether this attribute can be used as a query
+            constraint.
+          example: true
+        constraint_name:
```
CaseyTa, line 783: I think the constraint name is supposed to be the same original_attribute_name, so this may be redundant.

vdancik: Not necessarily as the same attribute can have different original names from different sources as shown in an example below.

vdancik:
Example meta-attribute JSON:

```
{
    "attributes": [
        {
            "attribute_type_id": "biolink:p_value",
            "attribute_source": "infores:ctrp",
            "original_attribute_names": [
                "p-value"
            ],
            "constraint_use": true,
            "constraint_name": "p-value"
        },
        {
            "attribute_type_id": "biolink:p_value",
            "attribute_source": "infores:chembl",
            "original_attribute_names": [
                "pValue"
            ],
            "constraint_use": true,
            "constraint_name": "p-value"
        },
        {
            "attribute_type_id": "biolink:publication",
            "attribute_source": "infores:chembl",
            "original_attribute_names": [
                "PubMed",
                "reference"
            ],
            "constraint_use": false
        }
    ]
}
```

edeutsch (approved):

rtroper (approved):

CaseyTa:
I still have a few questions:

1. The schema only defines `attribute_type_id` as required, but are we expected to provide as much detail as possible for all attributes (similar to the example above)? 
2. How does this work for the ARAs or KPs that may be accessing dynamic information (e.g., the Knowledge Collaboratory)?  
3. Does the presence of `constraint_name` imply that clients should specify that `constraint_name` exactly? Or would it still essentially be ignored by the service in favor of the `attribute_type_id`?

marcdubybroad (approved):

CaseyTa (approved):

PR #262, opened 2021-06-24
Addition of workflow template

edeutsch:
This PR adds the necessary components to define a common TRAPI workflow description as a list of steps.
Query.workflow lists the requested workflow steps
Response.workflow lists the workflow steps actually performed
Supported workflow step ids MUST be listed in the x-trapi info section
The full definition of workflow operation ids is contained at an external URL.


edeutsch:
An important open question is: Do we insist that all KPs implement the workflow explicitly for the "lookup" operation?
Or can it remain implicit by just sending a query graph?

CaseyTa (approved):

marcdubybroad (approved):

rtroper (approved):

patrickkwang (approved):

kshefchek (approved):

PR #266, opened 2021-07-01
Add tag for /meta_knowledge_graph

edeutsch:
Seems like it is customary to provide an entry in tags but we forgot in 1.1. Fix now?


CaseyTa (approved):

marcdubybroad (approved):

patrickkwang (approved):
I don't think that this is necessary or very informative, but it doesn't hurt.

kshefchek (approved):

rtroper (approved):

vdancik (approved):

PR #278, opened 2021-07-29
Update reference docs

patrickkwang:


edeutsch:
worth adding 'enum' support? for example for LogLevel

edeutsch (approved):

PR #283, opened 2021-08-17
Add forgotten explicit nullable: true attribributes

edeutsch:


TranslatorReasonerAPI.yaml
```diff
@@ -879,6 +886,7 @@ components:
             Node attributes provided by this TRAPI web service.
           items:
             $ref: '#/components/schemas/MetaAttribute'
+          nullable: true
```
patrickkwang, line 889: I would prefer a sensible default to `null`:
```suggestion
          default: []
```

TranslatorReasonerAPI.yaml
```diff
@@ -909,6 +917,7 @@ components:
             Edge attributes provided by this TRAPI web service.
           items:
             $ref: '#/components/schemas/MetaAttribute'
+          nullable: true
```
patrickkwang, line 920:
```suggestion
          default: []
```

TranslatorReasonerAPI.yaml
```diff
@@ -928,12 +937,14 @@ components:
...
             Source of an attribute provided by this TRAPI web service.
           example: infores:chembl
+          nullable: true
         original_attribute_names:
           type: array
           items:
             type: string
           description: >-
             Names of an the attribute as provided by the source.
+          nullable: true
```
patrickkwang, line 947:
```suggestion
          default: []
```

TranslatorReasonerAPI.yaml
```diff
@@ -997,6 +1009,7 @@ components:
             - '>'
             - <
             - matches
+          nullable: false
```
patrickkwang, line 1012: `operator` is required.
```suggestion
```

TranslatorReasonerAPI.yaml
```diff
@@ -1007,13 +1020,15 @@ components:
             A 'list' data type paired with the '>' or '<' operators will
             encode extraneous comparisons, but this is permitted as it is in
             SQL and other languages.
+          nullable: false
```
patrickkwang, line 1023:
```suggestion
```

patrickkwang (changes requested):

kshefchek (approved):

TranslatorReasonerAPI.yaml
patrickkwang: Per WG discussion, the empty array here actually means a different thing than null.

TranslatorReasonerAPI.yaml
patrickkwang: Per WG discussion, we prefer null to empty array here. Instead, let's add `minItems: 1`.

marcdubybroad (approved):

kshefchek (approved):

vdancik (approved):

rtroper (approved):

bill-baumgartner (approved):

TranslatorReasonerAPI.yaml
edeutsch: What did we decide to do here? Do we allow MetaEdge.attributes to be "nullable: true" or do we switch to "default: []"?
I recall some spirited discussion, but the conclusion does not appear to be captured in the minutes.
I vaguely recall that there were more proponents of "nullable: true", but it is very hazy.

patrickkwang: We want to allow both null and empty, to distinguish between "I'm not telling you" and "I'm telling you there's nothing".

PR #285, opened 2021-08-17
Fix documentation to reflect prior removal of 'relation'

edeutsch:


vdancik (approved):

kshefchek (approved):

marcdubybroad (approved):

vdancik (approved):

rtroper (approved):

bill-baumgartner (changes requested):
These change looks good. I suggest an additional minor change to the final sentence in the description: remove references to 'class' and 'subclass of', e.g.,  

> ... an exact match to the given QEdge predicate term, or to a term that is a descendant of the QEdge predicate term.

In OWL, the Biolink predicates that could belong to a QEdge are not referred to as *classes* but as *object properties* so it is potentially confusing to use 'class' and 'subclass of' here IMO.

edeutsch:
> 
> 
> These change looks good. I suggest an additional minor change to the final sentence in the description: remove references to 'class' and 'subclass of', e.g.,
> 
> > ... an exact match to the given QEdge predicate term, or to a term that is a descendant of the QEdge predicate term.
> 
> In OWL, the Biolink predicates that could belong to a QEdge are not referred to as _classes_ but as _object properties_ so it is potentially confusing to use 'class' and 'subclass of' here IMO.

yes, this seems reasonable. I have updated the PR to reflect this.

PR #287, opened 2021-08-19
add a limit to batch size

vdancik:
Add a limit to batch size to prevent runaway queries with large batches


patrickkwang:
Thanks, @vdancik! This is a discussion that needs to happen. The Ranking Agent team has at this point shifted entirely from a parallel-requests approach to a batched-requests approach, but we've observed that some KPs don't love receiving 1000+ CURIEs at once. My general feeling is that this is still the right approach, because it permits the KP all of the information/power to optimize performance at the web framework/handler level. If we divide them into multiple requests, on the other hand, they have to handle it at the server level, which seems typically harder.

I'm interested to hear from KPs about how the tradeoffs play out for them.

patrickkwang (changes requested):
What is the format of these limits?

* Does this batch limit apply to `ids` lists, `categories` lists, and `predicates` lists? What if you have multiple such lists in a query? Do we limit the sum of their lengths, the product of their lengths, ...?
* Rather than a "throughput" limit, I think we want a request rate limit? Rate limits can be implemented in a variety of ways. Does "3 requests per minute" mean that no request will be accepted less than 20 seconds after the last one? Or that a fourth request less than 60 seconds after the first one will be rejected? There are other ways, too...

ehinderer (approved):

patrickkwang:
Replaced by #294.

PR #294, opened 2021-09-09
Add batch-size and rate limits

patrickkwang:
Replaces #287.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -109,6 +109,22 @@ paths:
             application/json:
               schema:
                 type: string
+        '413':
+          description: >-
+            Payload too large. Indicates that batch size was over the limit
+            specified in x-trapi.
+          content:
+            application/json:
+              schema:
+                type: string
+        '429':
+          description: >-
+            Too many requests. Indicates that the client issued requests that
+            exceed the rate limit specified in x-trapi.
+          content:
+            application/json:
+              schema:
+                type: string
         '500':
           description: >-
             Internal server error.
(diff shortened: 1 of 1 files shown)


vdancik (approved):

kshefchek (approved):

marcdubybroad (approved):

CaseyTa:
IIRC, I believe we decided that rate limit would be expressed as requests/minute, and batch size limit would be a limit on the length of the longest IDs list. I think it would be helpful to add these to the descriptions.

edeutsch (approved):
This seems fine. I'm wondering if camelCase is the design we want? The only precedent in this area is biolink-version which is kebab-case. Seems like endpoints and properties all use snake_case. Classes use PascalCase.

Is there a rational motivation to use camelCase here for these?

CaseyTa:
Ignore my previous comment, I didn't realize these are described in x-trapi extensions definition: https://github.com/NCATSTranslator/translator_extensions/pull/11/files

CaseyTa (approved):

bill-baumgartner (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -25,6 +25,8 @@ info:
     asyncquery: REPLACE-WITH-true-IF-/asyncquery-IS-IMPLEMENTED-ELSE-false
     operations:
       - LIST-ALL-SUPPORTED-OPERATION-IDS-HERE-OR-REMOVE-THIS-SECTION
+    batch_size_limit: ADD-BATCH-SIZE-LIMIT-OR-REMOVE-IF-NO-LIMIT
+    rate_simit: ADD-REQUEST-RATE-LIMIT-OR-REMOVE-IF-NO-LIMIT
```
rtroper, line 29: Typo here (rate_simit)

patrickkwang: Haha, oops. Good catch, thanks.

rtroper (approved):

brettasmi (approved):

edeutsch (approved):

PR #306, opened 2021-12-01
Propose QEdge.exclude

edeutsch:


CaseyTa (approved):

kshefchek (approved):

marcdubybroad (approved):

bill-baumgartner (approved):

kennethmorton (approved):

vdancik (changes requested):
QEdge.exclude does not indicate what is excluded. I think QEdge.excluded_predicates is a better choice.

edeutsch:
> QEdge.exclude does not indicate what is excluded. I think QEdge.excluded_predicates is a better choice.

I don't understand what you mean that it does not indicate what is excluded. The documentation states "If set to true, then all subgraphs that could contain this edge are excluded from the final results."

It seems to me that QEdge.excluded_predicates is a different idea than what is proposed here. Consider:

A[disease] <- B[gene], "biolink:related_to"
C[small molecule] -> B[gene], "biolink:related_to"
C[small module] -> A[disease], exclude "biolink:contraindicated_for"

"Find all small molecules related to genes that are related to disease A, except for the small molecules (C) that are contraindicated for disease A"

I don' think this scenario can be expressed with QEdge.excluded_predicates, but can be with QEdge.exclude

I think these two proposals are not equivalent. I think:

`QEdge.excluded_predicates` should be simply thought of as `QEdge.omitted_predicates` from the `predicates` set
while
`QEdge.exclude` should be be thought of as `QEdge.exclude_subgraph_if_this_edge_is_possible`

colleenXu:
I'm wondering why we can't use predicate mixins to avoid specific predicates?

The logic here seems to be a "NOT predicate X". However, I think it's somewhat complicated to do this, especially since it involves navigating the hierarchy "NOT predicate X and its descendants". 

---
Also, a note for the example above:

BTE would consider the situation above a "cycle" (we disregard edge directions when making this decision). To find small molecules contraindicated for disease A, we'd probably want another node: Disease A -> Gene -> SmallMolecule <-(everything not contraindicated)- Disease A

kennethmorton:
From the Operations and Workflow group - It was asked if this could be addressed in a workflow.

Currently the answer is no, however a workflow operation could be developed to serve as a filter. After further discussion, it was decided that even with a suitable workflow operation, the group would recommend that this information also be contained in the query.

edeutsch:
Adding a link to this discussion, which is relevant:
https://github.com/RTXteam/RTX/issues/1766

PR #309, opened 2021-12-17
Propose QEdge.excluded_predicates

brettasmi:
This is a competing proposal to implement a restriction on the predicates that can be included in a result. The other proposal can be found here: https://github.com/NCATSTranslator/ReasonerAPI/pull/306

Here we introduce the `excluded_predicates` property on the `QEdge` object as an array of predicates that the server must exclude from any results. The naming of the property is certainly flexible, and we can change it if we want to make it shorter.

Note that ordering of the QEdge property list has changed such that the required `subject` and `object` fields appear before any optional properties. I think this is generally accepted OpenAPI style, and it makes auto-generated docs a little more friendly.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -594,12 +594,6 @@ components:
         an exact match to the given QEdge predicate term,
         or to a term that is a descendant of the QEdge predicate term.
       properties:
-        predicates:
-          type: array
-          items:
-            $ref: '#/components/schemas/BiolinkPredicate'
-          minItems: 1
-          nullable: true
         subject:
           type: string
           example: https://omim.org/entry/603903
@@ -614,6 +608,28 @@ components:
             Corresponds to the map key identifier of the
             object concept node anchoring the query filter
             pattern for the query relationship edge.
+        predicates:
+          type: array
+          items:
+            $ref: '#/components/schemas/BiolinkPredicate'
+          minItems: 1
+          nullable: true
+        excluded_predicates:
+          type: array
+          items:
+            $ref: '#/components/schemas/BiolinkPredicate'
+          minItems: 1
+          nullable: true
+          description: >-
+            An array of predicates that should be excluded from any results.
+            For example, a client may want to explore all possible relationships
+            between a drug and a disease but is not interested in
+            contraindications. In this case, the client would specify the
+            predicates field with ['biolink:related_to'] (or leave the
+            predicates field unspecified) and would specify the
+            excluded_predicates field with ['biolink:contraindicated_for'].
+            In this case, the server must exclude any results that have the
+            contraindicated_for predicate.
         constraints:
           type: array
           description: >-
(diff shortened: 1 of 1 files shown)


brettasmi:
@vgardner-renci , can you tag the review crew?

edeutsch:
One advantage of the #306 (other) proposal is that it allows something like this:

A[disease] <- B[gene], "biolink:related_to"
C[small molecule] -> B[gene], "biolink:related_to"
C[small module] -> A[disease], exclude "biolink:contraindicated_for"

"Get me all small molecules related to genes that are related to disease A, except for the small molecules that are contraindicated for disease A"

I don't you can express the same scenario with the current proposal?

CaseyTa:
@brettasmi, The first part of the description sounds a bit like only the `contraindicated_for` predicate should be excluded, as opposed to the overall result. For example, say a server has in it's KG:
A -[related_to]-> B
A -[contraindicated_for]-> B
The description sounds like the server should return the `related_to` edge and not return an edge with `contraindicated_for`, as opposed to not returning the `related_to` edge because the `contraindicated_for` edge exists. Or is this the intention of the alternate proposal? 

Also, how should constraints be handled? Let's say that in some QGraph, predicates, excluded_predicates, and constraints are all specified. Do we assume that the constraints are only applied to the predicates and not to the excluded_predicates?

vdancik (approved):

brettasmi:
@edeutsch 
> I don't you can express the same scenario with the current proposal?
I agree, that is an advantage of the other proposal. Would the proposed QGraph in your example require that there be some edge _other than_ `contraindicated_for` between A and C in order to return?

@CaseyTa 
I'm not sure I completely follow your message here, but the intention is that the server would only exclude `contraindicated_for` (and any descendants) relationships from the answer graph and any other descendant of `related_to` would be fair game. In other words, the set difference of(related_to  + descendants) and (contraindicated_for + descendants) could be included in the results. Does that make sense? If so, does it answer your question? 

As for constraints, I'd say that they should only apply to predicates that could possibly be returned.

CaseyTa:
@brettasmi Yes, thank you, that is how I interpreted it. But with this understanding, I believe your proposal achieves a different purpose than #306, as opposed to a competing proposal. I believe this proposal allows a query to essentially ignore certain types of edges if they exist, whereas #306 allows a query to exclude certain answers if a particular edge exists. Each proposal can achieve certain things the other can't. They could potentially be considered separately.

CaseyTa (approved):

edeutsch:
> @brettasmi Yes, thank you, that is how I interpreted it. But with this understanding, I believe your proposal achieves a different purpose than #306, as opposed to a competing proposal. I believe this proposal allows a query to essentially ignore certain types of edges if they exist, whereas #306 allows a query to exclude certain answers if a particular edge exists. Each proposal can achieve certain things the other can't. They could potentially be considered separately.

agreed!

brettasmi:
Closing per 2022-01-13 discussion:
- this functionality is achievable via careful specification of the `predicates` field
- this does not replicate the functionality of #306

PR #312, opened 2022-02-03
Add Attributes to NodeBindings

cbizon:
TRAPI 1.2 has attributes on EdgeBindings, but not on NodeBindings.  This proposes extending the attributes to NodeBindings.

The use case for ARAGORN is when we add unasked-for nodes in response to a TRAPI query (usually as a result of result merging), we want to record why those particular nodes got added to a particular result.


edeutsch (approved):
Looks fine to me. Maybe replace "nodee" with "node".

marcdubybroad (approved):

CaseyTa (approved):

PR #315, opened 2022-02-25
Remove SubAttribute and allow unlimited descending attributes

edeutsch:
Reverse the limit of allow just two levels of attributes. Now allow as many levels as desired.


marcdubybroad (approved):

CaseyTa (approved):
The schema change looks fine, though I believe we're waiting on decisions from Data Modeling Committee before this gets merged, is that correct?

vdancik (approved):

edeutsch:
> The schema change looks fine, though I believe we're waiting on decisions from Data Modeling Committee before this gets merged, is that correct?

Correct, this is a draft for what the current proposal is. We will discuss further before merging. And we can always change it again even after merging.

PR #326, opened 2022-05-05
Alternative support for qualifiers

edeutsch:
Alternative to #325, based on discussion in #325

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -784,10 +784,44 @@ components:
           items:
             $ref: '#/components/schemas/Attribute'
           nullable: true
+        qualifiers:
+          type: array
+          description: >-
+            A list of qualifiers that provide additional detailed nuance
+            to the assertion beyond what the subject-predicate-object can
+            provide.
+          items:
+            $ref: '#/components/schemas/Qualifier'
+          nullable: true
       additionalProperties: false
       required:
         - subject
         - object
+    Qualifier:
+      description: >-
+        An additional nuance attached to an assertion
+      type: object
+      properties:
+        type:
+          type: string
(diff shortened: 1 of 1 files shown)


vdancik (changes requested):
Can we be more specific, e.g. type => qualifier_type and value => qualifier_value

vdancik (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -784,10 +784,44 @@ components:
...
+          type: string
+          description: >-
+            The value associated with the type of the qualifier, drawn from
+            a set of controlled values by the type as specified in
+            the Biolink model (e.g. 'decreased' or 'increased' for the type
+            subject_direction, etc).
+          example: decreased
+          nullable: false
+      required:
+        - type
```
mbrush, line 823: should this read 'qualifier_type_id'

TranslatorReasonerAPI.yaml
```diff
@@ -784,10 +784,44 @@ components:
...
+            object_aspect, object_direction, etc).
+          example: subject_aspect
+          nullable: false
+        value:
+          type: string
+          description: >-
+            The value associated with the type of the qualifier, drawn from
+            a set of controlled values by the type as specified in
+            the Biolink model (e.g. 'decreased' or 'increased' for the type
+            subject_direction, etc).
```
mbrush, line 819: Consider making the example here consistent with the example of a qualifier_type_id above (where you used 'subject_aspect').  So here, say (e.g. 'expression' or 'abundance' for the qualifier type 'subject_aspect', etc).

edeutsch:
Closing this proposal in favor of #330

PR #327, opened 2022-05-16
Add QEdge.qualifier_constraints and rename to attribute_constraints

edeutsch:


brettasmi (approved):

vdancik:
What would be a mechanism for OR and NOT queries, e.g get me all compounds that decrease activity or synthesis of gene but not its expression?

ehinderer:
@vdancik That query has "biolink:decreases_expression_of" in the edge predicate list, isn't the point to exclude that logic? Also some of those predicates don't seem to be in the Biolink model.

edeutsch:
Closing because subsumed into #330

PR #330, opened 2022-05-21
A more complete Qualifiers implementation

edeutsch:


TranslatorReasonerAPI.yaml
```diff
@@ -632,15 +632,27 @@ components:
...
-        constraints:
+        attribute_constraints:
           type: array
           description: >-
-            A list of contraints applied to a query edge.
+            A list of attribute contraints applied to a query edge.
             If there are multiple items, they must all be true (equivalent
             to AND)
           items:
             $ref: '#/components/schemas/QueryConstraint'
```
mbrush, line 642: Now that we have the ability to constrain queries based on values of Attributes and Qualifiers, consider renaming the QueryConstraint object 'AttributeConstraint' or 'AttributeQueryConstraint'?  And then name the object that defines a constraint based on a Qualifier value 'QualifierConstraint' or 'QualifierQueryConstraint'.  This would make naming more clear and consistent.

TranslatorReasonerAPI.yaml
```diff
@@ -632,15 +632,27 @@ components:
...
+          type: array
+          description: >-
+            A list of QualifierSets that provide nuance to the QEdge.
+            If multiple QualifierSets are provided, there is an OR
+            relationship between them. If the QEdge has multiple
+            predicates, then each QualifierSet SHOULD define the
+            predicate it is qualifying via the
+            QualiferSet.qualified_predicate property.
+          items:
+            $ref: '#/components/schemas/QualifierSet'
```
mbrush, line 654: As noted above, consider renaming 'QualifierSet' --> 'QualifierConstraint' or 'QualifierQueryConstraint', to be more clear and consistent with the name of the object that holds a constraint definition based on Attribute values. 

Also, this avoids issue that a 'QualifierSet' object as presently defined doesn’t actually hold a set of Qualifiers.  It holds one Qualifier (which is comprised of a key-value pair of fields), along with optional predicate it is linked to.

TranslatorReasonerAPI.yaml
```diff
@@ -784,10 +796,69 @@ components:
...
+            slots in the Biolink model (e.g. subject_aspect, subject_direction,
+            object_aspect, object_direction, etc).
+          example: subject_aspect
+          nullable: false
+        value:
+          type: string
+          description: >-
+            The value associated with the type of the qualifier, drawn from
+            a set of controlled values by the type as specified in
+            the Biolink model (e.g. 'decreased' or 'increased' for the type
```
mbrush, line 831: Consider making the example here consistent with the example of a qualifier_type_id above (where you used 'subject_aspect'). So here, say "(e.g. 'expression' or 'abundance' for the qualifier type 'subject_aspect', etc)." And then change the example below from 'decreased' to 'expression'.

TranslatorReasonerAPI.yaml
```diff
@@ -784,10 +796,69 @@ components:
...
+            subject_direction, etc).
+          example: decreased
+          nullable: false
+      required:
+        - type
```
mbrush, line 836: Should 'type' here read 'qualifier_type_id'?

TranslatorReasonerAPI.yaml
```diff
@@ -784,10 +796,69 @@ components:
...
+        - value
+    QualifierSet:
```
mbrush, line 838: As noted above, consider renaming 'QualifierConstraint'

TranslatorReasonerAPI.yaml
```diff
@@ -784,10 +796,69 @@ components:
...
+      additionalProperties: false
+      description: >-
+        A set of Qualifiers that act together to add nuance to an assertion.
```
mbrush, line 841: With the renaming/reframing of this object, update definition to something like: 
"Defines a query constraint based on the type and value of a Qualifier attached to an edge.   e.g. can constrain a "ChemicalX - affects ?Gene" query to return only edges where ChemicalX affects the *expression* of some Gene, by specifying a qualifier_type of "object_aspect", and value of "expression".

TranslatorReasonerAPI.yaml
```diff
@@ -784,10 +796,69 @@ components:
...
+        qualifiers:
+          type: array
+          description: >-
+            A list of qualifiers that together provide additional detailed
+            nuance to the assertion beyond what the subject-predicate-object
+            can provide.
+          items:
+            $ref: '#/components/schemas/Qualifier'
+          nullable: false
+        qualified_predicate:
```
mbrush, line 852: 'qualified_predicate' is the name of an edge property defined in Biolink.  May be confusing to have a TRAPI property with the same name. Consider 'constrained_predicate'?

TranslatorReasonerAPI.yaml
```diff
@@ -784,10 +796,69 @@ components:
...
+          nullable: false
+      required:
+        - type
+        - value
+    QualifierSet:
+      additionalProperties: false
+      description: >-
+        A set of Qualifiers that act together to add nuance to an assertion.
+      properties:
+        qualifiers:
```
mbrush, line 843: I would consider defining the 'QualifierConstraint' object in a bit simpler way - as a object with three properties:
- **qualifier_type_id**: string (1..1)
- **value**: string (1..1)
- **constrained_predicate**:  BiolinkPredicate (0..m)

I don't think there is a need to or specific benefit in referencing / reusing the Qualifier object here - given that we may want to associate a given qualifier with >1 predicate in the constraint definition.

TranslatorReasonerAPI.yaml
```diff
@@ -784,10 +796,69 @@ components:
...
+            can provide.
+          items:
+            $ref: '#/components/schemas/Qualifier'
+          nullable: false
+        qualified_predicate:
+          $ref: '#/components/schemas/BiolinkPredicate'
+          description: >-
+            The predicate that is to be qualified in cases where there
+            are multiple predicates in a query and there thus might be
+            ambiguity.
```
mbrush, line 857: In this initial draft of an object to support Qualifier-based query constraints, I would also include a boolean 'not' property, as exists for the QueryConstraint object used for Attribute-based constraints . Key for the basic functionality of excluding specific qualifiers from a result.

TranslatorReasonerAPI.yaml
vdancik: I think we can use `AttributeConstraint` and `QualifierConstraint` to distinguish the two kinds of constrains.

TranslatorReasonerAPI.yaml
edeutsch: good idea, fixed.

TranslatorReasonerAPI.yaml
edeutsch: Qualifier is used both for QEdges and Edges. So it used both for querying and for conveying assertions. That is why the definition is slanted more toward the assertion side. In the context of queries, it specifies the constraints on the desired assertions. So okay as is?

TranslatorReasonerAPI.yaml
edeutsch: I was initially skeptical, but now agree that 'qualifier_set' may perhaps be a little better.

TranslatorReasonerAPI.yaml
edeutsch: I am neutral, but fine to change if it seems less conflicting with Biolink.

edeutsch:
So this is where I see the current PR:
image

and this is what @mbrush is suggesting:
image

I am fine with these changes except for the "not". That might need some more discussion.

TranslatorReasonerAPI.yaml
mbrush: True that 'Qualifier' object is used in asserting knowledge and querying for it.  But the definition here is for a 'QualifierSet' object, which we agreed above to rename 'QualifierConstraint' (because its express purpose is to allow constraining queries based on the values of Qualifiers on an edge).  
So given that we are not defining a  'QualifierConstraint' object, it makes sense to me to update the definition so that it clearly conveys this purpose of this object in constraining queries.

TranslatorReasonerAPI.yaml
mbrush: This suggestion was motivated by an example query Valdo provided that asks "What compounds increase the activity or phosphorylation of PPARA but **not** its abundance".  But as you point out, negation/exclusion are tricky and perhaps it is best to leave this out for now, until we look into it more closely.

TranslatorReasonerAPI.yaml
edeutsch: After pondering further, I'm convinced that QualifierConstraint is better. I have updated the PR.

TranslatorReasonerAPI.yaml
edeutsch: After pondering further, I'm convinced that qualifier_set is better. I have updated the PR.

TranslatorReasonerAPI.yaml
edeutsch: I understand better now and after pondering further, I'm convinced that constrained_predicate is better. I have updated the PR.

TranslatorReasonerAPI.yaml
edeutsch: I understand now. Yes, agreed. Updated with minor edits. See if the committed version seems agreeable.

edeutsch:
okay, everyone, after updating the PR based on @mbrush excellent comments, we now have this structure in the PR:
image

Please everyone examine carefully and review. This is our release candidate for qualifier support in TRAPI.

edeutsch:
Thanks for the review. I think your definition adjustments all seem sensible and I will make those unless there are objections. But I would like to discuss the property name a bit further:

> change the name of the Edge property qualifier_set back to qualifiers (to be consistent with the name of the attributes property here)

It is true that you requested in the previous round to change QualifierConstraint.qualifiers to QualifierConstraint.qualifier_set, and then I went ahead and also changed Edge.qualifiers to Edge.qualifier_set even though you didn't suggest it. I did that for what I perceived as consistency. So I am surprised that you would have us change it back for consistency.

It seems to make more sense to me to keep both QualifierConstraint.qualifier_set and Edge.qualifier_set consistent since they are essentially the same thing. rather than aim for consistency between Edge.attributes and Edge.qualifier[_set|s].

I thought calling it qualifier_set rather made a lot of sense because the Qualifiers within a set really do work together closely. Whereas the attributes are mostly a pile of independent things (some are provenance, some are inherent properties, some are assigned attributes, etc.)

So I rather like qualifier_set. But I'm happy to be outvoted.

How about all readers that made it this far put in their votes?
- Rocket for Edge.qualifier_set
- Heart for Edge.qualifiers

mbrush:
I am fine with either name for the Edge property.  I don't feel like we _need_ to use the same property name in the Edge and QualifierConstraint objects, since the referenced/nested Qualifier objects are serving different purposes in these different objects/contexts.  I have a slight preference for `Edge.qualifiers`, but am fine with `Edge.qualifier_set` . . . esp if we update the `QualifierConstraint.qualifier_set` definition as proposed so it is clear what role Qualifiers are playing here.  @edeutsch were you ok with the proposed changes to this definition?

edeutsch:
> edeutsch were you ok with the proposed changes to this definition?

@mbrush Yes, I think they're great. I was leaving a little time for others to chime in, but hearing none, the latest commit to the PR now uses your definitions. Seem good?

brettasmi (approved):

mbrush:
@edeutsch looks good to me!  Still slight preference for the name Edge.qualifiers over Edge.qualifier_set . . . but if vote remains 1-1 we can keep it as is.  Thanks for your work on this!

mbrush (approved):

kennethmorton (approved):

edeutsch:
okay, based on today's TRAPI call and follow-up with @mbrush, I have made the following changes to the PR:

Removed constrained_predicate
https://github.com/NCATSTranslator/ReasonerAPI/pull/330/commits/ef1a94f17bb8c392c61f19e24f4a80d9324cdc18

Rename Edge.qualifier_set to Edge.qualifier
https://github.com/NCATSTranslator/ReasonerAPI/pull/330/commits/4d527286d78c3cc4ba41484b1a354a029a12508a

Did I do this correctly?
any further concerns?
If correct, we are aiming to merge later today. Please respond.

TranslatorReasonerAPI.yaml
```diff
@@ -784,10 +797,67 @@ components:
...
+          type: array
+          description: >-
+            A set of Qualifiers that serves to add nuance to a query,
+            by constraining allowed values held by Qualifiers
+            on queried Edges.
+          items:
+            $ref: '#/components/schemas/Qualifier'
+          nullable: false
+      required:
+        - qualifiers
```
vdancik, line 859: @edeutsch , shouldn't `required` field in `QualifierConstraint` contain `qualifier_set` instead of `qualifiers` ?

edeutsch: yes, absolutely, good catch! Fix in PR with https://github.com/NCATSTranslator/ReasonerAPI/pull/330/commits/f4bcfc532d80e990b30bfc8294c8fecd2e7e5f1d
thanks!

vdancik (approved):

PR #341, opened 2022-06-28
Add x-translator.infores attribute

edeutsch:
This may have already been implemented by most everyone, but not formally listed here in the TRAPI template, so here it is.


colleenXu (approved):
see https://github.com/NCATSTranslator/translator_extensions/blob/123cd2422e812990029046f7025fc62630e1d0ce/x-translator/smartapi_x-translator_schema.json#L19 for the actual requirements

vdancik (approved):

newgene (approved):

PR #362, opened 2022-07-29
More complicated qualifier examples + rules documentation

sierra-moxon:
- [x] add two more example files, one for subject and object qualifiers, and one for a complex qualifier statement using GO-CAMs.
- [x] added a first pass at some documentation to go with the examples
- [x] added a small bit of test code so that the examples themselves will be validated against the current specification on PR.  Along with this addition, I added a requirements.txt file to make running the tests locally easier.  I am happy to remove this test if it is too heavyweight for this repo.  

I am still working on refining the examples, and the predicates team is adding additional rules, but I wanted to get early feedback if this is the right direction or if it is too much for this repo and should be stored elsewhere.


edeutsch:
This is really great, thank you!

Can you articulate why it seems preferable to make qualified predicate a top level property rather than part of the qualifier_set? It would seem okay in either place on Edges. But for QEdges, it can't go at the top because there might be multiple predicates and multiple qualifier_sets? (although probably rare)

edeutsch (approved):
This looks great to me and is welcome in this repo. I like it. My only question is if
```
                "qualifier_type_id": "biolink:qualified_predicate",
                "qualifier_value": "causes"
```
should perhaps be:
```
                "qualifier_type_id": "biolink:qualified_predicate",
                "qualifier_value": "biolink:causes"
```

PR #365, opened 2022-08-17
Update reference to Workflow Operations Schema 1.3.2

edeutsch:
This is potentially pretty disruptive, depending on how this is implemented. But yet it would seem like a major oversight to release TRAPI 1.3 still pointing to Workflow Operations Schema 1.0.0 when the latest is 1.3.2

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -236,7 +236,7 @@ components:
         workflow:
           description: List of workflow steps to be executed.
           allOf:
-            - $ref: http://standards.ncats.io/workflow/1.0.0/schema
+            - $ref: http://standards.ncats.io/workflow/1.3.2/schema
           nullable: true
         submitter:
           type: string
@@ -284,7 +284,7 @@ components:
         workflow:
           description: List of workflow steps to be executed.
           allOf:
-            - $ref: http://standards.ncats.io/workflow/1.0.0/schema
+            - $ref: http://standards.ncats.io/workflow/1.3.2/schema
           nullable: true
         submitter:
           type: string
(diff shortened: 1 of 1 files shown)


CaseyTa (approved):

colleenXu:
I brought this up, since I had been advised by members of the Operations/Workflow Working Group to use `lookup_and_score` rather than `lookup` for BTE/Service Provider, to describe what these tools are doing. We recently made a change to list `lookup_and_score` as the operation we did in the workflow section of our Responses (this change is not deployed on all instances yet - but it is in the development instance that also has TRAPI 1.3 code). 

We then noticed that our Responses were failing TRAPI 1.2 validation. See the discussion here (internal Translator Slack [link](https://ncatstranslator.slack.com/archives/C013Q5TVC87/p1660692160958049?thread_ts=1660681259.430349&cid=C013Q5TVC87)). 

---

notes:
* the translator_extension's README uses this link https://standards.ncats.io/operation.json, where a version is not specified (maybe the latest is used?). 
* However, the translator_extension's enum for operations has to be manually edited as the operations list changes. https://github.com/NCATSTranslator/translator_extensions/blob/2c6e200deb57aa45d2468035036e2fad34e91bf6/x-trapi/smartapi_x-trapi_schema.json#L26

cbizon:
It is potentially disruptive, but IMO better to fix it now rather than let it fester.

cbizon (approved):

kennethmorton (approved):

marcdubybroad (approved):

dkoslicki (approved):

vdancik (approved):

Issue #366, opened 2022-08-17
Add `nullable: true` to NodeBinding.query_id


gaurav:
The documentation of NodeBinding.query_id says that this is "An optional property" and that under certain circumstances "this query_id MUST be null or absent". I think it would be useful to add a `nullable: true` to the definition to make it clear that this field can be null.

https://github.com/NCATSTranslator/ReasonerAPI/blob/7520ac564e63289dffe092d4c7affd6db4ba22f1/TranslatorReasonerAPI.yaml#L481-L492

edeutsch:
This issue was addressed in PR #367, closing.

PR #367, opened 2022-08-18
explicitly declare query_id as nullable: true

edeutsch:
per suggestion #366


gaurav (approved):
Looks good to me! Thanks, Eric!

vdancik (approved):

PR #375, opened 2022-09-23
`abstract`, `mixin` and `deprecated` `category` and `predicate` Biolink Model terms

RichardBruskiewich:
Clarify documentation concerning use of `abstract`, `mixin` and `deprecated` `category` and `predicate` Biolink Model terms in TRAPI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 The Translator Reasoner API (TRAPI) defines a standard HTTP API for communicating biomedical questions and answers. It leverages the [Biolink model](https://github.com/biolink/biolink-model/) to precisely describe the semantics of biological entities and relationships. TRAPI's graph-based query-knowledge-binding structure enables expressive yet concise description of biomedical questions and answers.
 
-TRAPI is described primarily by an [OpenAPI](https://github.com/OAI/OpenAPI-Specification) document [here](TranslatorReasonerAPI.yaml). The request/response structure is also documented in a more human-readable form [here](docs/reference.md).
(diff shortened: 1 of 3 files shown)


edeutsch (changes requested):
I think reference.md is autogenerated from the yaml, so any changes to reference.md will be lost. These documentation changes need to be made to yaml. and we should first create a 1.4 branch and start working in there before we merge.

RichardBruskiewich:
Hmm.. OK. @edeutsch, I'll take another run at this.

RichardBruskiewich:
OK, @edeutsch, I manually transferred the comments that I manually included in the **reference.md**, into the **TranslatorReasonerAPI.yaml** file. I'm leaving those revisions in **reference.md**, since they clarify semantics but don't change the function of the schema in TRAPI 1.3 (no harm in telling folks now, rather than later, right?). The next release cycle will simply overwrite the **reference.md** with essentially the identical comments (just without some of the fancy markdown highlights in the curated version).

edeutsch (changes requested):
Would you rebase this to the "1.4" branch?

RichardBruskiewich:
> Would you rebase this to the "1.4" branch?

I've tried doing that. I'm all thumbs with esoteric Git comments, though, so I'm not sure if it worked.

edeutsch:
I'm thinking that if successful, instead of this:
image

we should see that you want to merge into 1.4 instead of `master`

?

If you can't get it to work, I suppose since it's just a documentation change, we could merge it into master with little harm.

edeutsch (approved):

vdancik (approved):

edeutsch (approved):

edgargaticaCU (approved):

vdancik (approved):

marcdubybroad (approved):

kennethmorton (approved):

MarkDWilliams (approved):

PR #377, opened 2022-10-04
Make predicate required and cleanup

edeutsch:
Make predicate required and not nullable and clean up examples, definitions, and make the allOf to oneOf switch for validator compatibility

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -818,39 +818,48 @@ components:
         resulting from a query upon the underlying knowledge source.
       properties:
         predicate:
-          allOf:
+          description: >-
+            The type of relationship between the subject and object
+            for the statement expressed in an Edge.
+          example: biolink:gene_associated_with_condition
+          oneOf:
             - $ref: '#/components/schemas/BiolinkPredicate'
-          nullable: true
+          nullable: false
         subject:
-          $ref: '#/components/schemas/CURIE'
-          example: OMIM:603903
           description: >-
             Corresponds to the map key CURIE of the
             subject concept node of this relationship edge.
+          example: MONDO:0011382
+          oneOf:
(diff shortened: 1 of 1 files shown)


edgargaticaCU (approved):

vdancik (approved):

RichardBruskiewich (approved):

ehinderer (approved):

andrewsu (approved):

marcdubybroad:
Just to verify, does this mean that this query would be invalid (no predicate)?
```
  {
    "message": {
      "query_graph": {
        "edges": {
          "e00": {
            "subject": "n01",
            "object": "n00"
          }
        },
        "nodes": {
          "n00": {
            "categories": ["biolink:Gene"]
          },
          "n01": {
            "ids": ["MONDO:0011936"],
            "categories": ["biolink:DiseaseOrPhenotypicFeature"]
          }
        }
      }
    }
  }
```

edeutsch:
Correct, that would become a schema violation. Is that okay? Or do we want to support such loose queries?
If one simply added a predicate biolink:related_to, that would have the same meaning I suppose, and would be legal under the proposed scheme.

marcdubybroad:
Should we also require categories then on nodes? Just trying to stay consistent on the rules for the json query.

edeutsch:
I think the general sentiment was that if you specify a CURIE, then a category was either redundant or potentially conflicting. Many felt that if you provide a CURIE, you should not provide a category. But there was some disagreement as well.

MarkDWilliams (approved):

vdancik:
@marcdubybroad actually the PR only specifies that a **knowledge-graph** edge is required; predicate in QEdge remains optional and your query example will still be valid.

marcdubybroad (approved):

RichardBruskiewich (approved):

PR #402, opened 2023-03-15
Update ChangeLog.md

RichardBruskiewich:


edeutsch (changes requested):
should be in the "not yet merged" section

edeutsch (approved):

PR #307, opened 2021-12-01
Propose QEdge.negated and Edge.negated

edeutsch:
Addresses #281

PR #407, opened 2023-03-21
Introduce a special property to make "negation" of edges more visible in a TRAPI message

sierra-moxon:
We discussed in the modeling group how to represent a 'negated' edge in our models either as a negated form of each of our predicates, e.g. `predicate: not_affects` and `predicate: affects` or as an edge property `negation`.  Ultimately, the DM group decided to include a `negation` edge property ([revisit the discussion and vote here](https://github.com/biolink/biolink-model/discussions/826)), with the caveat that `negation` should _never_ be ignored downstream and should _always_ be shown.  Right now in TRAPI, the `negation` edge property would be treated in TRAPI like any other `attribute`.  

Groups making negated edges still express concern that the negation property will be ignored or less visible than necessary for groups downstream (and ultimately users).   This PR attempts to give more transparency to the `negation` property via a specific property on an edge, as well as adding a flag to MetaEdge so that a resource an alert users that an edge might have both positive and negated edges.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -1072,6 +1072,15 @@ components:
           oneOf:
             - $ref: '#/components/schemas/CURIE'
           nullable: false
+        negated:
+          description: >-
+            Indicates whether the statement expressed in an Edge is
+            negated. If true, the statement is negated.  This also 
+            reflects the implementation in Biolink Model for 
+            negating an edge.
+          type: boolean
+          nullable: false
+          default: false
         attributes:
           description: A list of additional attributes for this edge
           items:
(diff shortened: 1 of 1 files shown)


edeutsch:
See also PR #307, which is a little different. That PR includes the idea of adding negated to QueryEdge, too, which probably should be part of it?

sierra-moxon:
Thank you for the pointer;  I missed the QEdge. :)  I merged that PR branch into this one, but happy to go the other way instead if that makes more sense.

edeutsch (commented):
thanks for merging, this looks good to me. I will close #307 in favor of this, but note that there is some good discussion over there that should be acknowledged.

sierra-moxon:
Link to @cmungall's comments on negation of an edge from PR #307 : https://github.com/NCATSTranslator/ReasonerAPI/pull/307#issuecomment-1034084848

Issue #414, opened 2023-03-31
Should we use the fancy supporting graph infrastructure for subclass inference?


edeutsch:
There was some musing about whether simple subclass inference as performed by KPs should also use the auxiliary / supporting graph infrastructure.

This seemed appealing to some. But others felt that there was no need to take something that is already working and change it so substantially. We are undecided. This issue is a placeholder to consider this further and make a decision.

This issue illustrates the ongoing conversation: https://github.com/NCATSTranslator/Feedback/issues/117

colleenXu:
Our team is still thinking this through, and we have asked for help in the reasonerapi Slack channel. I'm pasting my post from that channel below

---

Topic 1: how to represent a result when ID-expansion is done?

AKA an entity-curie on a QNode was expanded to include ontology-descendants. In the [migration guide](https://github.com/NCATSTranslator/ReasonerAPI/blob/1.4/MigrationAndImplementationGuide1-4.md?plain=1#L261), we saw that `query_graph` cannot be modified, and `result.node_bindings` must only include QNodes from `query_graph`.

Q1: Is it still okay to use result.node_bindings.query_id to represent that ID-expansion was done? See my example here 

Q2: If the answer to Q1 is no, then what is the recommended way to represent the info in my example? I walk through that same info here, trying to use TRAPI 1.4 auxiliary graphs + creating edges...but I'm not sure that this is correct. It also seems kinda complicated / hard to parse. 

Q3: If the answer to Q2 is to use auxiliary graphs + create edges ... then is "nesting" okay? 
* AKA one edge `e0` references an auxiliary graph `gA` (in its attributes). `gA` is a list of edges, one of which is `e5`.  `e5` also references an auxiliary graph, `gZ`
* we ask because if we use auxiliary graphs/creating-edges for ID-expansion AND for collapsing multi-hop paths from creative-mode, this is a situation that could happen...

colleenXu:
And now for my impressions (not representing my teams as a whole):

A. implementing `result.node_bindings.query_id` should be much easier for our team (we haven't done it yet because we have been [waiting for more guidance](https://github.com/biothings/BioThings_Explorer_TRAPI/issues/497#issuecomment-1242482638))

B. while our team could implement an auxiliary-graph / creating-edges method, I'm not enthusiastic about it...

* it's more complicated and wordy to use, based on the walk-through I did above...(affecting memory/storage negatively?)
[...]

PR #420, opened 2023-04-04
Update Migration Guide

uhbrar:
This adds a guide dedicated more to KP's as opposed to ARA's. This also includes an example for source retrieval provenance refactoring.


edeutsch:
You can more easily read the new formatted guide here: https://github.com/uhbrar/ReasonerAPI/blob/update_guide/MigrationAndImplementationGuide1-4.md

edeutsch (commented):
There seemed to be some substantial push-back against implementing this policy when we brought this up, so I think this will need some more discussion before merging.

MigrationAndImplementationGuide1-4.md
```diff
@@ -893,4 +893,166 @@ Now we can put these two sections back together to get the final message.
...
+                        "id": "diabetes"
+                    }
+                ],
+                "query_node_1": [
+                    {
+                        "id": "metformin"
+                    }
+                ]
+            },
+            "edge_bindings": {
```
edeutsch, line 1048: I don't think results have edge_bindings any more?

edeutsch (changes requested):

MigrationAndImplementationGuide1-4.md
uhbrar: Resolved

uhbrar:
I thought that it was pretty well accepted that subclasses could be represented this way when discussed at architecture. There was definitely discussion about this use case being added to the guide.

edeutsch:
My recollection from the TRAPI call is that this was discussed and there was substantial push-back, perhaps from Vlado and others, I don't recall now exactly. The notes say:
```
There was mention of KPs potentially using AuxiliaryGraph for subclass encoding? Do we want to pursue that? Come up with a recommendation/example on how to encode a subclass reasoning response from a KP? Point people: Chris, Abrar, Matt, Andrew?
For now, the policy is to just keep doing it the current way, and we will consider a change in the future
[Eric] will start a thread to discuss if it is sensible to create a subclass example. DONE
```

I started the issue here: https://github.com/NCATSTranslator/ReasonerAPI/issues/414
There is some discussion, but I suspect others will want to contribute.

I think this change will need formal approvals before merging.
There are multiple revisions in this PR so it is hard to see the specific proposed changed, but this link may be more helpful:
https://github.com/NCATSTranslator/ReasonerAPI/pull/420/commits/e9c41d59a8e9e30f28f9491b4bbb970b80bae8d6

I suggest conversion continue in the issue.
And the topic could be revisited in the Architecture and/or TRAPI calls next week.

MigrationAndImplementationGuide1-4.md
```diff
@@ -1,10 +1,12 @@
-# TRAPI 1.4 Migration and Implementation Guide
+This guide is split into two major sections. The first section focusses on ARA's. The second section focusses on KP's.
```
colleenXu, line 1: focusses -> focuses
Also apostrophe use? For plural, probably use "ARAs", "KPs", and "CURIEs".

MigrationAndImplementationGuide1-4.md
```diff
@@ -225,7 +334,7 @@ Most of the knowledge graph will remain the same in this example. Only one of th
     "predicate": "treats",
     "attributes": [
         {
-            "attribute_type_id": "support graph",
+            "attribute_type_id": "infores:support_graphs",
```
colleenXu, line 337: is this supposed to be "biolink:support_graphs", from [here](https://github.com/biolink/biolink-model/blob/35925815e6ef47c0e4a0e28417452be486c1c18f/biolink-model.yaml#L672)?

MigrationAndImplementationGuide1-4.md
```diff
@@ -452,7 +569,7 @@ Finally, we can put all of it back together, to show the full message.
                 "predicate": "treats",
                 "attributes": [
                     {
-                        "attribute_type_id": "support graph",
+                        "attribute_type_id": "infores:support_graphs",
```
colleenXu, line 572: see https://github.com/NCATSTranslator/ReasonerAPI/pull/420/files#r1162416075

MigrationAndImplementationGuide1-4.md
```diff
@@ -522,4 +643,421 @@ Finally, we can put all of it back together, to show the full message.
...
+```
+
+## TRAPI 1.4 Example Message
+
+```
+"message": {
+    "query_graph": {
+        "nodes": {
+            "query_node_0": {
+                "ids": ["MONDO:0005148"]
```
colleenXu, line 978: should use MONDO:0005015 (plain diabetes m.). This curie is actually for type2...

MigrationAndImplementationGuide1-4.md
colleenXu: I recall in the Architecture call 2 weeks ago, there was a question on whether KPs could use result.edge_bindings (particularly since they are often called to do one-hop edge-retrieval by ARAs). Was this question resolved? Does it depend on the subclassing discussion?

colleenXu:
Also does the description for Result need editing since it discusses edges and edge-bindings? https://github.com/uhbrar/ReasonerAPI/blob/1.4/TranslatorReasonerAPI.yaml#L564

MigrationAndImplementationGuide1-4.md
uhbrar: Fixed. Thank you.

There's noting in the specification that necessarily says that it has to be a list, but I think expected practice is a list as opposed to a single value. Unless this was further specified in the biolink model.

MigrationAndImplementationGuide1-4.md
uhbrar: Fixed.

MigrationAndImplementationGuide1-4.md
uhbrar: Fixed. Thank you.

uhbrar:
I didn't notice that. I thought I'd already fixed the description for Result. I'll put in a separate pull request in for that.

MigrationAndImplementationGuide1-4.md
```diff
@@ -1,10 +1,12 @@
-# TRAPI 1.4 Migration and Implementation Guide
+This guide is split into two major sections. The first section focuses on ARAs. The second section focusses on KPs.
```
colleenXu, line 1: still has a "focusses" typo

edeutsch (approved):

vdancik (approved):

edeutsch (approved):

edgargaticaCU (approved):

RichardBruskiewich (approved):

edeutsch:
The number of approvals is very meagre, but I'm going to merge this anyway, because if we actually wait for approvals we'll never merge it.

PR #421, opened 2023-04-04
Fix required fields

uhbrar:
I noticed in #416 and #418 that the "required" fields didn't have the updated field names. This fixes that.


edeutsch (approved):
thanks, good catch

PR #423, opened 2023-04-10
Fix example typo in Metaknowledge Graph Example JSON

sierra-moxon:


tokebe (approved):

colleenXu:
Okay, so each MetaQualifier in the qualifiers section of one MetaEdge should have a unique qualifier_type_id....is what I'm getting here. Is that right?

This may be less important, but I'm confused on what this example represents:
* the subject is ChemicalEntity, so does the aspect qualifier "expression" make sense here?
* also unclear if the qualified predicate makes sense in the qualified statement: abundance of a chemical "causes" activity of a Gene (vs taking it out).

edeutsch (commented):
This looks like a reasonable correction of an inadvertent duplication. But are the values really correct? Expression of a chemical entity? Abundance of a gene? should the subject and object perhaps have somewhat different applicable values?

vdancik (approved):

PR #430, opened 2023-04-24
add sources field to MetaEdge so that infores identifiers are returne…

sierra-moxon:
…d in the meta knowledge graph


edeutsch (commented):
I think this is a reasonable addition, but I question if we can tolerate to make a mandatory (and thus breaking) addition at this late stage of 1.4. I would be happier with making it optional in 1.4 and mandatory in 1.5 is we so desire.

edeutsch (approved):

edeutsch (approved):

vdancik (approved):

edgargaticaCU (approved):

RichardBruskiewich (approved):

PR #431, opened 2023-04-27
Updates to the implementation guide

sierra-moxon:


examples/Attributes/publications.md
```diff
@@ -33,30 +23,72 @@ publications:
   range: publication
 ```
 
-Note, the range for `biolink:publications` is `biolink:publication` which is actually a [Biolink Model class](https://biolink.github.io/biolink-model/docs/Publication.html#class-publication).  However, using a collection of CURIEs as shorthand syntax for representing a full list of `biolink:Publication` objects 
-is expected in a TRAPI message (see below).
+### Implementation Guidance
 
+1. The `biolink:publications` edge property MUST be used as the `attribute_type_id` to report 
+publications that support an Edge.  Each reported edge MUST have at least one Attribute object with
```
mbrush, line 29: > "Each reported edge MUST have at least one Attribute object with `attribute_type_id`:`biolink:publications`" 

I don't think we want to require that every Edge MUST have a publication Attribute. There are many that just don't have one.

examples/Attributes/publications.md
```diff
@@ -92,22 +124,24 @@ as a list in the `value` field, as below:
...
-7. Where multiple types of publication identifiers or URLs exist for a single Publication (e.g. a PMID, PMCID, 
+4. A source MAY provide a URL as one of the values of the `biolink:publications` attribute, EXCEPT in cases where  
+a Pubmed, Pubmed Central, or DOI identifier is part of a full URL. In such cases, the identifier MUST be reported 
+in its CURIE form:  
+    a. http://www.ncbi.nlm.nih.gov/pubmed/29076384   →  **PMID: 29076384**  
+    b. http://europepmc.org/articles/PMC6246007      →  **PMC:6246007**  
+    c. https://doi.org/10.1080/17512433.2018.1398644 →  **DOI:0.1080/17512433.2018.1398644**  
+
+
+5. Where multiple types of publication CURIEs exist for a single Publication (e.g. a PMID, PMCID,
```
mbrush, line 139: We should clarify the intent of this instruction. I had a couple things in mind:
 
1. If a source provides multiple identifiers for the same pub, which I think some will do - KPs should report only one of them - in order of precedence: PMID, PMCID, DOI, bespoke identifier.

2. If a source provides only a PMCID, or DOI - do we instruct the KP to try to find a PMID for the publication and report this where it exists?

sierra-moxon: @cbizon @gaurav  - does NodeNorm return a normalized identifier for Publication?

cbizon: No, it doesn't, though I suppose it could if that were useful....

PR #437, opened 2023-06-17
Add nullable: true

edeutsch:
There's quite a lot of TRAPI 1.4 in the wild with the following being null:
Message.auxiliary_graphs
RetrievalSource.upstream_resource_ids
RetrievalSource.source_record_urls

I think we intended them to be nullable. but the validator throws an error when these are null.
so let's update the 1.4.0-beta4 schema to truly make them nullable, making a lot of TRAPI 1.4 out (more) valid.


vdancik (approved):

uhbrar (approved):

RichardBruskiewich (commented):
approved

PR #441, opened 2023-07-01
Revert  auxiliary_graphs to original 1.4.0 plus nullable: true

RichardBruskiewich:
Validates cleanly, as  discussed


edeutsch (approved):
Thanks, yes, this is the desired end state

PR #449, opened 2023-08-10
Update retrieval_provenance_specification.md

mbrush:
Minor but helpful clarifications to the spec.


edeutsch (commented):
Seems pretty good. Could be a little tidier. How about:
- fix spelling of Retreival
- Fix "Information Resource" to "InformationResource" to me consistent
- Remove double space in "in the  Edge"
- Maybe `Edge.source` property "contains" rather than "points to"?
- superfluous comma in source_record_urls?

Aside from these trivial things, it all seems good to me

mbrush:
Yes, thanks Eric.  Just fixed all these!

edeutsch (approved):
looks great, thanks

PR #473, opened 2024-01-04
Proposed bypass_cache directive in Query

edeutsch:


vdancik (approved):

edgargaticaCU (approved):

cbizon:
Just following up from the 2/13 Architecture call - is there a definition of how much caching needs to be bypassed to comply?   Just to muddy the waters with some examples - 

ARAGORN caches KP edges, and then on top of that caches full responses.  In a bypass-cache scenario is the goal to bypass both kinds?

Unsecret "caches" in the sense that they pre-build a KG based on large inputs that they periodically receive. Since there is no way to rebuild that on the fly, any downstream caching doesn't really matter since the source data won't change.

I think ARAX pre-caches MVP1 results in a way that can't really be regenerated on the fly, how should it respond? (Please correct if wrong)

So I guess maybe it's going to be bit fuzzy, but is the idea something like "bypass all caching as much as possible?"

edeutsch:
I have refined the definition of bypass_cache based on conversation to:

            Set to true in order to request that the agent obtain
            fresh information from its sources in all cases where
            it has a viable choice between requesting fresh information
            in real time and using cached information.

Please approve, or suggest further refinements, or comment with problems you have with this, or at least be prepared to come to a resolution at tomorrow's TRAPI call.

cbizon:
I like this refinement FWIW

gprice1129 (approved):

brettasmi (approved):
I'm approving, but I think the standard way to handle this is a `Cache-Control: no-cache` header, which tells the caching server to validate the cached response as "fresh" before responding. [Ref](https://datatracker.ietf.org/doc/html/rfc9111#section-5.2). I'm not sure this behavior maps exactly to what Translator is doing, but it's standard for bypassing caches.

Also note that allowing users to explicitly bypass the cache could be used maliciously overload the server, so it's advised that services implementing such behavior also implement some amount of rate limiting, etc.

vdancik (approved):

edgargaticaCU (approved):

PR #480, opened 2024-04-05
Update TranslatorReasonerAPI.yaml

mbrush:
added a Qnode set_id property, to support multi-curie query work.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -899,16 +899,18 @@ components:
             - MANY
           nullable: true
         set_id:
-          $ref: '#/components/schemas/CURIE'
           description: >-
-            A CURIE-based identifier for a set that is defined as the QNode. A   
-            set id MUST be created by the system that defines the queried set, 
-            whenever the set_interpretation value is 'MANY' or 'ALL'. A set id 
+            A CURIE-based identifier for a QNode that is defined as a set. A   
+            set_id MUST be created by the system that defines the queried set, 
+            whenever the set_interpretation value is 'MANY' or 'ALL'. A set_id 
             MUST NOT be created when the set_interpretation value is 'BATCH'. 
             Any form/syntax and method for minting curie-based ids for sets can 
             be used, including UUIDs or bespoke prefixes and strings. Downstream
-            systems MUST re-use the original set ids as possible in the messages  
+            systems MUST re-use the original set_id as possible in the messages  
             they create/send, to facilitate any required merging operations.
+          nullable: true
+          oneOf:
+            - $ref: '#/components/schemas/CURIE'
         constraints:
           type: array
           description: >-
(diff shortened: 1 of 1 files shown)


mbrush:
To discuss on 4-11-24 TRAPI Call

edeutsch (commented):
In OpenAPI, it is my understanding that by defining set_id like this:
```
        set_id:
```
the CURIE is copied into set_id, wiping out any other properties (like description). And I think we want cardinality to be explicit. How about this:
```
        set_id:
          description: >-
            A CURIE-based identifier for a QNode that is defined as a set. A   
            set_id MUST be created by the system that defines the queried set, 
            whenever the set_interpretation value is 'MANY' or 'ALL'. A set_id 
            MUST NOT be created when the set_interpretation value is 'BATCH'. 
            Any form/syntax and method for minting curie-based ids for sets can 
            be used, including UUIDs or bespoke prefixes and strings. Downstream
            systems MUST re-use the original set_id as possible in the messages  
            they create/send, to facilitate any required merging operations.
          nullable: true
          oneOf:
            - $ref: '#/components/schemas/CURIE'
```

edeutsch (approved):

RichardBruskiewich (approved):

cbizon:
@uhbrar made what I thought was a good suggestion - instead of adding a set_id and using that to hold the UUID, instead make the UUID the id of the set node and instead add a member_ids list to hold the identifiers of the members of the set.  The reason I prefer that suggestion is because the results node bindings are more in line with the query that way (the qnode is bound to the set id, which is the id in the qgraph, just like any other bound node).

mbrush:
@cbizon What would happen in the case of a BATCH query under Abrar's proposal. There is no set identifier created in this case.  Would the list of queried CURIEs go into the `ids` field in this case, and the `member_ids' field is empty?  Contrast this to the case of a MANY or ALL set interpretation, where the `ids` field holds the set identifier, and the list of member CURIEs goes into the `member_ids` field.

cbizon:
@mbrush That's right.  And that variation in the qgraph is clearly the downside of that approach.  

But IMO it makes the result binding across the board cleaner, because the id defined in the query graph is what is bound in the result. (when it's batch the bindings are to phenoX, and when it's MANY or ALL, the binding is to the UUID, i.e. it's always to the `id` defined in the qgraph)

colleenXu:
I misinterpreted Chris Bizon/Abrar's proposal to be the following, and I thought it was an interesting idea so I'll describe it here...

What about making the set-id the QNode's label? So rather than "n00"/"n01", it's:

```
query_graph: {
    nodes: {
   	 n00: {
   		 categories: [biolink:Disease]
   	 },
   	 "uuid:a2414cb2314d235146": {
   		 ids: [hp:pheno1, hp:pheno2, hp:pheno3],
   		 categories: [biolink:PhenotypicFeature],
                 set_interpretation: MANY
   	 },
```

then result node-bindings would use the node labels provided (which include the set id)

edeutsch:
The conclusion from today's TRAPI call is that this PR is obsolete and replaced by #481.
Closing this one.

PR #481, opened 2024-04-10
Set id Update TranslatorReasonerAPI.yaml

mbrush:
Alternate proposal for representing set ids and set members in a Qnode, per the suggestion [here](https://github.com/NCATSTranslator/ReasonerAPI/pull/480#issuecomment-2047599410).

Note that the benefit of this over the proposal in #480 is that node binding in Results can continue to work as they currently do - where q nodes here are bound to curies in the `ids` field and not some new field like `set_id`.


vdancik (approved):

CaseyTa (approved):

edeutsch (approved):

CaseyTa (approved):

edeutsch:
Merging after discussion at TRAPI call today.

PR #484, opened 2024-05-10
The value of the  items property in the member_ids property should NOT be a YAML list!

RichardBruskiewich:
The value of the  **`items`** property in the **`member_ids`** property should NOT be a YAML list!

```yaml
        member_ids:
          type: array
          # etc...
          items:
            $ref: '#/components/schemas/CURIE'   # <= this is a 'scalar' argument to **`items`**, not a list
```


edeutsch:
replaced by #485. closing

PR #491, opened 2024-07-23
fix indent for MetaQualifier - applicable_values - example

bill-baumgartner:
The example for MetaQualifier - applicable_values was indented once too far such that it appeared under MetaQualifier - applicable_values - items instead. This commit corrects the indent level.


edeutsch:
Replaced by #531

PR #496, opened 2025-04-24
Add support for PathFinder extensions

edeutsch:


TranslatorReasonerAPI.yaml
```diff
@@ -1008,6 +1043,51 @@ components:
...
+          type: array
+        attribute_constraints:
+          default: []
+          description: A list of attribute contraints applied to a query edge. If
+            there are multiple items, they must all be true (equivalent to AND)
+          items:
+            $ref: '#/components/schemas/AttributeConstraint'
+          type: array
+      required:
+      - intermediate_nodes
```
uhbrar, line 1085: This might not be a required field. We only need intermediate nodes on constrained queries, so not all pathfinder queries require them.

edeutsch: The default is [ ], an empty list. I idea is that there should not be two ways to say there are no intermediate nodes: null or [ ]. Rather there should be one way: [ ]. Thus we make it required and require an empty list to say there are none. This is a custom that we have been moving to in other parts of the TRAPI schema, although I don't feel strongly.

colleenXu: The required empty lists to say "none" bloat the TRAPI response without adding any data. I think requiring properties when their value is "null" also bloats the TRAPI response. 

So I think this custom should be changed to having optional properties only - so tools don't need to provide the property if it's not required and they don't use it (aka remove from `required` section, don't use `nullable`).

tokebe: I tend to agree with @colleenXu ; It's a slight bit of extra consideration when parsing the TRAPI, but the bloat reduction is real. As long as everyone is playing on the same assumptions (which are given by descriptions in the TRAPI spec), I don't see a problem with it.

colleenXu: I agree that this discussion of required empty lists should be moved to a diff ticket and shouldn't be a barrier to getting this PR done. 

(I do wonder if it can be added to 1.6, if we are all in agreement and it's an easy, backwards-compatible change to make...)

vdancik (approved):

vdancik (approved):

vdancik (approved):

tokebe (approved):

colleenXu (approved):

colleenXu:
Info on the commits I just made:
* I removed PathBinding's attributes - during today's meeting, it sounded like Pathfinder group agreed "No one needs binding attributes so we can remove them". @uhbrar @edeutsch Please confirm that this is okay, and revert/change if needed. 
* pure formatting (no change to actual contents)

colleenXu:
I approved, but noticed two potential issues @uhbrar @edeutsch:

1. this spec has no restrictions on how many objects can be specified within `paths` for 1 Pathfinder `query_graph` (line 925-). Is that okay? 
   * I imagine only 1 object is expected in `paths`...and then, maxProperties can be used to limit the number of objects within `paths`. 
2. QPath predicates are currently nullable (line 1112) but also required (line 1125). Is that okay? 
   * I think this situation was raised as a problem/addressed earlier today in a different part of the schema.

ShervinAbd92 (approved):

TranslatorReasonerAPI.yaml
```diff
@@ -627,7 +628,9 @@ components:
             The list of all Analysis components that contribute to the result.
             See below for Analysis components.
           items:
-            $ref: '#/components/schemas/Analysis'
+            oneOf:
+              - $ref: '#/components/schemas/Analysis'
+              - $ref: '#/components/schemas/PathfinderAnalysis'
```
gprice1129, line 633: Is there any way we can tie together these oneOf choices so that a developer is forced to pick between the different versions? It seems like the way its specified you could mix using Analysis with PathfinderQueryGraph for instance.

TranslatorReasonerAPI.yaml
gprice1129, line 1: Some of the field language need to be updated where `X` was change to `BaseX` and `PathfinderX` was added. All of these looks like they were copy/pasted so they descriptions and other fields still refer to `X`.

TranslatorReasonerAPI.yaml
edeutsch: I guess it would have to be done higher up the chain. So we could have a Message and PathfinderMessage, such that PathfinderMessages would only have PathfinderQueryGraphs and PathfinderAnalysis. Such that one XMessage is committed to one or the other.

TranslatorReasonerAPI.yaml
edeutsch: Abrar is planning description refinements in the next week.

TranslatorReasonerAPI.yaml
edeutsch: Consensus from 2025-05-15 call: consensus is that it would be best to leave the schema as proposed because of future intended changes. Handle the risk of improper use with the semantic validator which can check that it is not used in a unintended way (e.g. combining Analysis and PathfinderAnalysis).

gprice1129 (approved):

PR #497, opened 2025-04-28
Pathfinder proposal edits

colleenXu:
my proposed solutions for Analysis/QueryGraph that are backward-compatible and add the Pathfinder features. The branch also has edits for formatting, so new sections match the rest of the document

I've verified that the schema works as-expected using https://smart-api.info/editor to review the "parsed" schema.

Notes: I used allOf and oneOf, as described in the OpenAPI's documentation on [Inheritance/model composition](https://swagger.io/docs/specification/v3_0/data-models/inheritance-and-polymorphism/) and [oneOf](https://swagger.io/docs/specification/v3_0/data-models/oneof-anyof-allof-not/).

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -716,14 +716,14 @@ components:
             items:
               $ref: '#/components/schemas/EdgeBinding'
         path_bindings:
-          additionalProperties:
-            items:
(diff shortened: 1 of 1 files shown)


colleenXu:
Another note: QEdge, QPath could also be adjusted to reduce repetitiveness (have a common base schema, then use allOf to add custom properties).

PR #499, opened 2025-05-22
Pathfinder guide

uhbrar:
Adds a specification guide for pathfinder queries, and also refines descriptions for pathfinder schema components.


uhbrar:
Erroneously merging into master instead of 1.6 branch.

PR #502, opened 2025-05-28
Option A for flattening path_bindings

colleenXu:
Removes PathBinding class, makes path_bindings a list of strings.


colleenXu:
Won't be adding to TRAPI 1.6-beta right now

PR #503, opened 2025-05-28
Option B for flattening path_bindings

colleenXu:
Adjusts PathBinding class, makes path_bindings a list of strings.


colleenXu:
Won't be adding to TRAPI 1.6-beta right now

PR #530, opened 2025-12-04
Add COLLATE to set_interpretation

edeutsch:
Clarify behavior of set_interpretation and member_ids.


edeutsch:
This assumes that if there are no ids, then then only option in COLLATE. If there is a distinction between set_interpretation=COLLATE and set_interpretation=null or absent, then this needs to be revised.

Did people envision

edeutsch:
Replaced with #543

PR #536, opened 2026-01-08
Top-level Edge KL/AT handling

tokebe:
> [!note]
> Depends on #533 

This PR extracts KL/AT from edge-attributes into top-level values on edges, and adds KL/AT constraints in the new constraints format.

PR #537, opened 2026-01-08
Top-level query parameters

tokebe:
> [!note]
> Depends on #533 

- Refactors AsyncQuery to inherit from Query
- Adds a top-level `parameters` field to Query and Response
- Migrates some Query fields to `parameters` and defines `timeout` parameter

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -121,6 +121,15 @@ paths:
             application/json:
               schema:
                 type: string
+        '409':
+          description: >-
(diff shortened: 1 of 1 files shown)


edeutsch (approved):

colleenXu:
Oops! Merging this PR "reverted" all the changes from the last two merged PRs #535 #536 

I think this could have been avoided if we...
1. merged/updated this branch with the 2.0 branch at the time (which had those merged PRs)
2. then double-checked the "Files changed" to make sure only our additions are included (and not any unexpected "revert" behavior)
3. and then merged this PR.

Issue #517, opened 2025-09-26
Flatten `node_bindings` into a list of strings (node IDs)


colleenXu:
[Depends on #506 AND #516, similar to what's being proposed in #505 for path_bindings]

The `node_bindings` values (defined using additionalProperties) are arrays of NodeBinding objects. **If we decide to remove `attributes` and `query_id` from `NodeBinding`** (see #506 AND #516), this object will only have 1 defined property, `id`.

If we aren't taking advantage of the NodeBinding object structure (having other properties besides `id`), **it would be more concise and simple to make the `node_bindings` value an array of strings (node ids) instead.**

---

The current spec

```
        node_bindings:
          type: object
          description: >-
            The dictionary of Input Query Graph to Result Knowledge Graph node
            bindings where the dictionary keys are the key identifiers of the
            Query Graph nodes and the associated values of those keys are
            instances of NodeBinding schema type (see below). This value is an
            array of NodeBindings since a given query node may have multiple
            knowledge graph Node bindings in the result.
          additionalProperties:
            type: array
            items:
              $ref: '#/components/schemas/NodeBinding'
            minItems: 1
          nullable: false 
...
    NodeBinding:
      type: object
      description: >-
          An instance of NodeBinding is a single KnowledgeGraph Node mapping,
          identified by the corresponding 'id' object key identifier of the
          Node within the Knowledge Graph. Instances of NodeBinding may
          include extra annotation in the form of additional properties.
          (such annotation is not yet fully standardized). Each Node
          Binding must bind directly to node in the original Query Graph.
      properties:
        id:
          oneOf:
            - $ref: '#/components/schemas/CURIE'
          nullable: false
          description: >-
            The CURIE of a Node within the Knowledge Graph.
        query_id:
          oneOf:
            - $ref: '#/components/schemas/CURIE'
          description: >-
            An optional property to provide the CURIE in the QueryGraph to
            which this binding applies. If the bound QNode does not have an
            an 'id' property or if it is empty, then this query_id MUST be
            null or absent. If the bound QNode has one or more CURIEs
            as an 'id' and this NodeBinding's 'id' refers to a QNode 'id'
            in a manner where the CURIEs are different (typically due to
            the NodeBinding.id being a descendant of a QNode.id), then
            this query_id MUST be provided. In other cases, there is no
            ambiguity, and this query_id SHOULD NOT be provided.
          nullable: true
        attributes:
          type: array
          description: >-
[...]

Issue #515, opened 2025-09-26
Flatten `edge_bindings` into a list of strings (edge IDs)


colleenXu:
[Depends on #506, similar to what's being proposed in #505 for path_bindings]

The `edge_bindings` values (defined using additionalProperties) are arrays of EdgeBinding objects. **If we decide to remove `attributes` from `EdgeBinding`** (see #506), this object will only have 1 defined property, `id`.

If we aren't taking advantage of the EdgeBinding object structure (having other properties besides `id`), **it would be more concise and simple to make the `edge_bindings` value an array of strings (edge ids) instead.**

---

The current spec

```
            edge_bindings:
              type: object
              description: >-
                The dictionary of input Query Graph to Knowledge Graph edge
                bindings where the dictionary keys are the key identifiers of
                the Query Graph edges and the associated values of those keys
                are instances of EdgeBinding schema type (see below). This
                value is an array of EdgeBindings since a given query edge may
                resolve to multiple Knowledge Graph Edges.
              additionalProperties:
                type: array
                items:
                  $ref: '#/components/schemas/EdgeBinding'
...
    EdgeBinding:
      type: object
      description: >-
        A instance of EdgeBinding is a single KnowledgeGraph Edge mapping,
        identified by the corresponding 'id' object key identifier of the
        Edge within the Knowledge Graph. Instances of EdgeBinding may include
        extra annotation (such annotation is not yet fully standardized).
        Edge bindings are captured within a specific reasoner's Analysis
        object because the Edges in the Knowledge Graph that get bound to
        the input Query Graph may differ between reasoners.
      properties:
        id:
          type: string
          description: The key identifier of a specific KnowledgeGraph Edge.
          nullable: false
        attributes:
          type: array
          description: >-
            A list of attributes providing further information about the
            edge binding. This is not intended for capturing edge attributes
            and should only be used for properties that vary from result to
            result.
          items:
            $ref: '#/components/schemas/Attribute'
          minItems: 1
          nullable: false
      additionalProperties: true
      required:
        - id
```

Issue #505, opened 2025-05-29
Flatten `path_bindings` into a list of string IDs


colleenXu:
Right now in 1.6-beta, `path_bindings` dictionary values are lists of PathBinding objects (defined using `additionalProperties`). But if we aren't taking advantage of the PathBinding object structure (having other properties besides `id`), **it would be more concise and simple** for `path_bindings` to be a list of aux graph ids (strings). 

**This issue is for us to revisit this idea in the future.** 

Previous ideas for how to change the schema to a list of aux graph ids: 
* https://github.com/NCATSTranslator/ReasonerAPI/pull/502: **removes** PathBinding class
* https://github.com/NCATSTranslator/ReasonerAPI/pull/503: **adjusts** PathBinding class

---

Current 1.6-beta spec:

```
            path_bindings:
              type: object
              description: >-
                The dictionary of input Query Graph paths to Analysis paths, specifically only for pathfinder queries.
              additionalProperties:
                type: array
                items:
                  $ref: '#/components/schemas/PathBinding'
...
    PathBinding:
      type: object
      description: >-
        A instance of PathBinding is a single binding of an input QueryGraph path
        (the key to this object) with the AuxiliaryGraph id containing a list of
        edges in the path. The Auxiliary Graph does not convey any order of edges
        in the path.
      properties:
        id:
          type: string
          description: The key identifier of a specific auxiliary graph.
      additionalProperties: true
      required:
        - id
```

colleenXu:
Addressed by #546, merged to 2.0 branch.

PR #546, opened 2026-02-12
Simplify binding structure

tokebe:
This simplification changes bindings so they all change from looking like this:

```json
...
"<node/edge/path>_bindings": {
          "key1": [
            {
              "id": "<CURIE/EdgeID/AuxGraphID>"
            }
          ],
          "key2": [
            {
              "id": "<CURIE/EdgeID/AuxGraphID>"
            }
          ]
        },
...
```

to this:

```json
...
"<node/edge/path>_bindings": {
          "key1": {
            "ids": ["<CURIE/EdgeID/AuxGraphID>"]
          }
          "key2": {
            "ids": ["<CURIE/EdgeID/AuxGraphID>"]
          }
        },
...
```

The reasoning for keeping the binding objects, rather than making it a simple `id: array`, is to leave room for possible future changes to `set_interpretation` response formatting. A likely-enough future change would be to use unified binding objects so attributes can be attached that reference the set of nodes/aggregation of edges (instead of the current format of using reasoner-constructed Nodes and Edges with support graphs). This format allows for that possibility, as well as testing prior to future TRAPI changes because the binding objects allow for arbitrary additional properties.

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -605,7 +605,7 @@ components:
             bindings where the dictionary keys are the key identifiers of the
             Query Graph nodes and the associated values of those keys are
             instances of NodeBinding schema type (see below). Because a given
-            QNode may have multple knowledge Nodes bound in the result,
+            QNode may have multiple knowledge Nodes bound in the result,
             the NodeBinding object may list multiple knowledge Nodes.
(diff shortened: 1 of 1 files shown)


edeutsch (approved):

tokebe:
Addresses #505, #515, #517

PR #547, opened 2026-02-19
Use anyOf instead of oneOf

tokebe:
Fix proposed by @colleenXu in 2026-02-19 meeting

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -712,10 +712,9 @@ components:
       additionalProperties: true
       required:
         - resource_id
-      oneOf:
+      anyOf:
         - required: [edge_bindings]
         - required: [path_bindings]
-        - required: [edge_bindings, path_bindings]
     EdgeBinding:
       type: object
       description: >-


edeutsch (approved):

colleenXu:
How I uncovered the issue:

* using #544 as a starting point, [made a json-schema of Analysis (plus all dependent subschemas)](https://github.com/user-attachments/files/25426234/_Analysis_schema.json)
* [example Analysis to adjust and see what validates or doesn't](https://github.com/user-attachments/files/25426232/_Analysis_example.json)
* plug into https://www.jsonschemavalidator.net/

The validation fails because having both edge_bindings and path_bindings fulfills >1 of the oneOf "required" sub-schemas.

---

For solutions:
* only removing line 718 (`required: [edge_bindings, path_bindings]`) was not sufficient.
* changing to [anyOf](https://json-schema.org/understanding-json-schema/reference/combining#anyOf) does what we want:
  * fulfilling >1 "required" sub-schema is now fine (having both types of bindings)
  * don't need line 718

PR #548, opened 2026-02-19
Add minProperties to QueryGraph.nodes

tokebe:
Proposed by @colleenXu in 2026-02-19 meeting, we don't want queries with no nodes to be valid

### DIFF
--- a/TranslatorReasonerAPI.yaml
+++ b/TranslatorReasonerAPI.yaml
@@ -816,6 +816,7 @@ components:
             on bound nodes.
           additionalProperties:
             $ref: '#/components/schemas/QNode'
+          minProperties: 1
         edges:
           type: object
           description: >-


vdancik (approved):

MarkDWilliams (approved):

colleenXu:
How I uncovered the issue:

* using #544 as a starting point, [made a json-schema of QueryGraph (plus all dependent subschemas)](https://github.com/user-attachments/files/25426963/_QueryGraph_schema.json)
* [example QueryGraph to adjust and see what validates or doesn't](https://github.com/user-attachments/files/25426968/_QueryGraph_example.json)
* plug into https://www.jsonschemavalidator.net/

The schema-validation succeeds when `nodes` is {} (empty)...but this leads to strange QueryGraphs that don't seem doable.

Like `{nodes: {}}` or the example above with an edge and path but no nodes.

